Issue #449, opened 2023-07-16
There are larger files than GitHub allows in the history preventing some `git push` to forks


Mikaela:
### Describe the bug

There are large files in git preventing some of the pushes.

```
remote: error: Trace: 9437e82b47c162cf1367bfee87fed122b7a6ca95e023b47eee0de54a7f9a1331
remote: error: See https://gh.io/lfs for more information.
remote: error: File ios/Runner.ipa is 123.78 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: File ios/Runner.ipa is 123.78 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com/.
```

Originally reported by `@shootingstardragon:matrix.org` at https://matrix.to/#/%23fluffychat%3Amatrix.org/%24HC9Uu2lHvnfuPph2cL67tOttB1_0CkQnHVp1hNp-58k?via=midnightthoughts.space&via=matrix.org&via=tchncs.de 

### To Reproduce

1. Clone or fork FluffyChat
2. `git remote add upstream-gitlab https://gitlab.com/famedly/fluffychat.git`
3. `git fetch --all`
4. `git push origin --tags`

### Expected behavior

The push goes through.

### Screenshots

```
remote: Resolving deltas: 100% (25028/25028), completed with 35 local objects.
remote: error: Trace: 355440c3121b86b4a54e730655dc7d32feb86ceafbf67a05d71c7f85aacd1de0                 
remote: error: See https://gh.io/lfs for more information.                                    
remote: error: File ios/Runner.ipa is 123.78 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: File ios/Runner.ipa is 123.78 MB; this exceeds GitHub's file size limit of 100.00 MB
remote: error: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.                                                                                                     
Transferred: sent 326009572, received 30824 bytes, in 88.5 seconds     
Bytes per second: sent 3682410.6, received 348.2                                                            
To github.com:Mikaela/fluffychat.git
 ! [remote rejected]     0.41.3 -> 0.41.3 (pre-receive hook declined)
...
```

### Platform (please complete the following information):

N/A

### Additional context

I fear this will require editing git history to resolve and force pushing.

[...]

PR #452, opened 2023-07-16
Add pre-commit and .gitattributes for small style fixes and EOL normalisation

Mikaela:
I was hoping the large file check would detect #449 and give instructions for proceeding, but while that didn't happen, this configuration hit a lot of missing EOFs, trailing whitespaces and one incident of executable `.desktop`.

The EOL normalisation is something I am just used to performing for my repositories as `eol=lf` while this just looks at which platform git is ran on and uses lf on Linux and macOS, crlf on Windows and I think internally everything is lf. A couple of files that appear rewritten were caught by `git add --renormalize .`.

In case this is desirable, the next step would be enabling https://pre-commit.ci for the repository


Mikaela:
At least the checks pass now

Mikaela:
I guess I should update this as there are new merge conflicts

Mikaela:
Or not as I have headache and there is no interest anyway.

PR #455, opened 2023-07-16
fix: Use the emoji picker's search in getSuggestions

### LINKED ISSUE #420: emoji auto complete is kinda broken
i think the new type name of emoji thing is kinda broken if i type `:rat` i don't find rat but if i type `:rodent` i do find rat i should be able to find the emoji by name instead of just category.

TheSunCat:
Fixes #420
Tested on Linux and Android.

### DIFF
--- a/lib/pages/chat/input_bar.dart
+++ b/lib/pages/chat/input_bar.dart
@@ -1,7 +1,7 @@
+import 'package:emoji_picker_flutter/emoji_picker_flutter.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
-import 'package:emojis/emoji.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:matrix/matrix.dart';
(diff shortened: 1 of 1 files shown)


TheOneWithTheBraid:
I really like the changes.

Two points to be considered :

- the renderer still uses `package:emojis`. In order to properly match all Emojis, it might be useful to sync the renderer using the other package too. E.g. `👩‍❤️‍👩 ` as a newer Emoji is simply recognized as regular text
- how far does that packages support skin tones ? In the drop down list I was only proposed yellow emojis

(Both should not block this PR but rather be discussed later.)

TheOneWithTheBraid:
Dart analyzer fails, can you please apply `dart format lib/ test/` and `dart run import_sorter:main --no-comments` ? Thanks !

TheSunCat:
> I really like the changes.

Thanks!

>     * the renderer still uses `package:emojis`. In order to properly match all Emojis, it might be useful to sync the renderer using the other package too. E.g. `👩‍❤️‍👩 ` as a newer Emoji is simply recognized as regular text

I don't know much about `package:emojis` (pretty new to Flutter) but it seems reasonable to use this package for everything emoji-related, so as to have consistent behavior througouht FluffyChat.

>     * how far does that packages support skin tones ? In the drop down list I was only proposed yellow emojis

This is the same package that was used for the emoji picker that comes up under the chat bar, which supports skin tones by long-tapping an emote. This is not really an option here as the suggestions list is awkward enough that long-tapping would be too much in my opinion.
An idea would be to support a global "default skin tone" option (maybe add a selector to the right of the emoji search bar from #454) to be used in this list. This would also allow faster access to sending an emote with the user's preferred skin tone from the emoji drawer, as they currently have to long-tap each emoji every time they need to send it.

PR #456, opened 2023-07-16
fix audio not stopping and returning to start after finishing

hermanncoders:


krille-chan (approved):

krille-chan:
Thank you so much for the fix. Unfortunately the formatting check failed. Can you run `dart format lib` in your branch to fix it?

krille-chan:
I've fixed it in another commit already. Thank you

PR #464, opened 2023-07-22
fix: code cannot be wrap

Decodetalkers:
I think it is a wrong fix for link, '\n' should not be replaceAll, so I think this should be reverted


krille-chan (approved):
Thanks. Lets try it out 😊

PR #468, opened 2023-07-24
Krille/import emoji improvements

krille-chan:


lib/pages/settings_emotes/settings_emotes.dart
```diff
@@ -343,32 +340,10 @@ class EmotesSettingsController extends State<EmotesSettings> {
...
-        if (kIsWeb || PlatformInfos.isMobile) {
-          await Share.shareXFiles(
-            [XFile(fileName, bytes: Uint8List.fromList(output))],
-          );
-        } else {
-          String? savePath = await FilePicker.platform
-              .saveFile(fileName: fileName, allowedExtensions: ['zip']);
-
-          if (savePath == null) {
-            // workaround for broken `xdg-desktop-portal-termfilechooser`
```
TheOneWithTheBraid, line 355: Can you please include this workaround then 🤔 ? It's otherwise completely broken for my use case 😂 ... Yes, I know, workarounding other packages bugs is always annoying but it's super useful to me.

lib/pages/settings_emotes/settings_emotes_view.dart
```diff
@@ -23,6 +25,31 @@ class EmotesSettingsView extends StatelessWidget {
       appBar: AppBar(
         leading: const BackButton(),
         title: Text(L10n.of(context)!.emoteSettings),
+        actions: [
```
TheOneWithTheBraid, line 28: I honestly prefer the prominent buttons but I can understand it's not anyone's preference. LGTM hence apart from the comment above.

lib/pages/settings_emotes/settings_emotes.dart
krille-chan: If this is broken, then also save all files is broken. So why fixing it at one place but not the other? Btw I could not reproduce it and I'm still able to save files under Linux.

TheOneWithTheBraid: You are likely not using `xdg-desktop-portal-termfilechooser` ... But yes. We should rather fix it in upstream, I will place a PR.

TheOneWithTheBraid:
Merged. Kan you review my PR in favor ?

PR #474, opened 2023-07-27
Scalable image preview for room avatars and profile images

Bnyro:
Hey,
this PR adds scalable (/zoomable) previews of room avatars and profile images, similar to how it's already possible for normal images that are sent in chats.
The image viewer can be found when clicking on a profile image in the user bottom sheet or when tapping the room avatar in the chat details.

Concerning the implementation: I had the feeling that the preview would share too much code with the existing `ImageViewer` for Matrix events containing images, thus they're currently both using the same component.
It'd be possible to create a new separate `ImageViewer` component only, but I'm not sure if the boilerplate code would be worth it.

And sorry for all the reformatting caused by the changes, unfortunately that seems to be necessary for the CI to pass.

### DIFF
--- a/lib/pages/chat/events/image_bubble.dart
+++ b/lib/pages/chat/events/image_bubble.dart
@@ -74,7 +74,7 @@ class ImageBubble extends StatelessWidget {
     showDialog(
       context: Matrix.of(context).navigatorContext,
       useRootNavigator: false,
-      builder: (_) => ImageViewer(event),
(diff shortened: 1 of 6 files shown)


Bnyro:
Just rebased on the latest changes, and I must say wow, great job with the redesign of the chat details!
It looks so amazing now, I'm already looking forward to finally being able to update the app via F-Droid :)

Bnyro:
No interest in continuing work on this.

PR #477, opened 2023-07-27
feat: restore latest filter on app start 

Bnyro:


krille-chan:
Hm, sorry I do not agree that we should make this configurable. This would be too much configuration option IMO. Instead we could just persist the last active filter and reopen the app exactly there when the app gets opened again

Bnyro:
I haven't been sure which solution you'd prefer, thus I decided to add it as an option since that's what most other apps do commonly.
However, I do agree with you, remembering the filter is probably the more user convenient solution.
Pushed the changes.

krille-chan:
Currently this does not work but I think this is the same problem as https://github.com/krille-chan/fluffychat/issues/614 I will try to solve it

krille-chan:
I have refactored the code base of the Store. Now it should be possible to do it as you have planned it @Bnyro  can you update your PR?

Bnyro:
I'm not currently able to build the app hence I can't tell if this still works unfortunately.

PR #478, opened 2023-07-28
small performance tweaks for Message

gilice:
It is the widget most often built, so every small bit helps.
- replace a List with an array (should be easier to optimize for the compiler)
- remove a Padding widget and use the parent Container's padding instead


krille-chan (approved):
LGTM 👍 thank you so much

Issue #169, opened 2023-07-15
Not recieving notifications on my second signed in account


krille-chan:
#### Description
I am unable to receive notifications on iPad OS on my second signed in account. Everything is enabled in both fluffychat's settings and the notifications settings in iPad OS. No devices show up under devices, neither in FluffyChat on iPad OS nor in Element on Desktop.

#### To Reproduce
Sign in to second account in fluffychat.
Have all notification settings enabled.
Receive no notifications.

#### Additional information:
- Device: iPad 5th generation
- OS and OS version: iPad OS 15.0.2 or 15.1
- Installed version of FluffyChat: 0.41.4, 0.42.0, 0.42.1, 0.42.2 or 1.0.1

mindmeshlink:
I'm signed into a couple accounts and running into this on Android and desktop Linux too. The current selected account gets notifications on desktop, and Android seems to only get them from the primary account.

xabirequejo:
Same issue here

- iOS: 17.2.1
- FluffyChat: 1.16.0

Notifications work on first account, but won't on second account.

viscountexx:
Still an active issue that makes the multi-account Fluffychat beta nigh-useless. I've resorted to using multiple separate instances of Nheko on desktop with the --profile flag to use different accounts.

xabirequejo:
I believe my notifications are working after the iOS 18.3 update.
Can anyone confirm?

AsenHu:
On Android 15, the problem persists. Specifically, fluffychat still does not set notification tags in accounts outside of the first account.

reox:
Seems like this was fixed recently? I started to get notifications for my second account now with Android 16 (Pixel 7) and FC 2.6.0

PR #522, opened 2023-08-17
fix: push on multiple clients 

wrenix:
- It is my first work with flutter be glade.
- Current the code is untested

---

https://gitlab.com/famedly/fluffychat/-/merge_requests/1126 again ...

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -30,8 +30,8 @@ void main() async {
 
   String? pin;
   if (PlatformInfos.isMobile) {
-    for(client in clients) {
-      BackgroundPush.clientOnly(clients);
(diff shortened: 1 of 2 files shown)


krille-chan:
Thanks but this solution does not work and also it does not scale good if you have N clients then every push message will be processed N times. We should try a different approach

wrenix:
1. Multiaccount is beta ... so if there was created higher ressource usage, i believe it is okay
2. why i do not let improve my code?  You have not make an review, you just criticie and block my code - that is not really cute

wrenix:
Related to:
  - #169 
  - #544 
  - #599

PR #524, opened 2023-08-18
chore: upgrade flutter to 3.13.0

selfisekai:


krille-chan:
Thanks but you have not fixed the lints before. Already done in https://github.com/krille-chan/fluffychat/pull/525

PR #541, opened 2023-08-23
feat/ChatListItem: small changes

gilice:
- extract hasNotifications to a variable
 this is more clear and potentially faster as we don't compute `notifications > 0` multiple times

- make a text style const


krille-chan (approved):
Looks valid to me. Thank you so much 👌

PR #545, opened 2023-08-26
fix: desktop drop of HTTP content on linux

TheOneWithTheBraid:
Meow, annoying bug.

### DIFF
--- a/lib/utils/desktop_drop_downloader.dart
+++ b/lib/utils/desktop_drop_downloader.dart
@@ -1,0 +1,48 @@
+import 'dart:io';
+
+import 'package:flutter/foundation.dart';
+
+import 'package:http/http.dart';
+import 'package:path_provider/path_provider.dart';
+
+abstract class DesktopDropDownloader {
+  const DesktopDropDownloader._();
+
+  static Future<String?> unsupportedUriCallback(String url) async {
+    if (kIsWeb) return null;
+    final uri = Uri.tryParse(url);
+    if (uri == null) return null;
+
+    if (!['http', 'https'].contains(uri.scheme)) return null;
+
+    Response response;
+
+    try {
+      response = await get(uri);
+    } catch (_) {
+      return null;
+    }
+
+    Directory tmp;
+
+    // that's likely failing on many distros but future proof for upcoming
+    // implementations
+    try {
+      tmp = await getTemporaryDirectory();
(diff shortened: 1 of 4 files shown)


pubspec.yaml
```diff
@@ -148,6 +148,11 @@ msix_config:
   install_certificate: false
 
 dependency_overrides:
+  desktop_drop:
+    git:
+      url: https://github.com/TheOneWithTheBraid/flutter-plugins.git
+      path: packages/desktop_drop
+      ref: braid/unsupported-uri-scheme-callback
```
krille-chan, line 155: dependency_overrides need a comment with a linked issue or MR when to solve this

lib/utils/desktop_drop_downloader.dart
```diff
@@ -0,0 +1,48 @@
...
+
+import 'package:flutter/foundation.dart';
+
+import 'package:http/http.dart';
+import 'package:path_provider/path_provider.dart';
+
+abstract class DesktopDropDownloader {
+  const DesktopDropDownloader._();
+
+  static Future<String?> unsupportedUriCallback(String url) async {
```
krille-chan, line 11: Is it possible to implement this in the desktop_drop package instead? It could be an optional feature there.

lib/utils/desktop_drop_downloader.dart
```diff
@@ -0,0 +1,48 @@
...
+    if (kIsWeb) return null;
+    final uri = Uri.tryParse(url);
+    if (uri == null) return null;
+
+    if (!['http', 'https'].contains(uri.scheme)) return null;
+
+    Response response;
+
+    try {
+      response = await get(uri);
```
krille-chan, line 21: This seems to have no user feedback. When I move an url to a 100GB file into it, will it just try to download 100GB into RAM without the GUI showing any feedback?

lib/utils/desktop_drop_downloader.dart
```diff
@@ -0,0 +1,48 @@
...
+    } catch (_) {
+      return null;
+    }
+
+    Directory tmp;
+
+    // that's likely failing on many distros but future proof for upcoming
+    // implementations
+    try {
+      tmp = await getTemporaryDirectory();
```
krille-chan, line 31: Could be done with platform checks instead of try-catch blocks

Issue #413, opened 2023-07-15
Metadata leak in voice messages


krille-chan:
#### Description
Voice message metadata is not stripped. Audio Metadata should be disabled by default, as it is on many other chatting applications(e.g. WhatsApp, Telegram clients), or at least the user should have the option to disable it.

#### To Reproduce
Send a voice message via FluffyChat, download it and probe it with ffprobe, you will see:
```
  Metadata:
    major_brand     : mp42
    minor_version   : 0
    compatible_brands: isommp42
    creation_time   : 2022-08-10T10:19:33.000000Z
    com.android.version: 12
    com.android.manufacturer: OnePlus
    com.android.model: ONEPLUS A6013
  Duration: 00:00:02.93, start: 0.000000, bitrate: 73 kb/s
    Stream #0:0(eng): Audio: aac (LC) (mp4a / 0x6134706D), 22050 Hz, mono, fltp, 64 kb/s (default)
    Metadata:
      creation_time   : 2022-08-10T10:19:33.000000Z
      handler_name    : SoundHandle
```

#### Additional information:
- Device: Oneplus 6T
- OS and OS version: LineageOS / Android 12
- Installed version of FluffyChat: 1.5.0
- (Android only) Which store are you using: F-Droid
- (Android only) Are Google Services available: No

CodeLuh:
Concerning this issue, is a byte manipulation based approach possible ?

I have checked the header of the `m4a` file format used in audio, and it seems like the concerned field containing all the metadata is (unsurprisingly) the one named `meta`. 
As described by the container, it consists of a 32 bits unsigned number specifying the length, followed by another 32 bits unsigned number called `meta`, then the actual metadata which is in the range. The first number also counts itself and the `meta` field.

After personal tests, stripping out the field and modifying the length specified before the `moov` field didn't work. 
However, nulling all of the content of the metadata (except the length field and name field) results in a playable file without any metadata shown by `ffprobe`.

After searching in the repos, the relevant manipulation probably happen as far as I know around [this variable](https://github.com/krille-chan/fluffychat/blob/2ca4c5b4bad415bb29dc5adba1828b833dff0493/lib/pages/chat/chat.dart#L621C27-L621C27).

Let me know your thoughts on this solution.

PR #548, opened 2023-08-30
fix: Metadata leak in voice message

TrivialDem:
This pull request addresses the issue of leaking some metadata on voice messages by setting it to 0. Note that it only works on `mp4` container and will do nothing otherwise.

Fixes [413](https://github.com/krille-chan/fluffychat/issues/413).


krille-chan:
Can you add some more information to this? What will happen with other formats than mp4? Actually I would prefer to have this as an option in the record package. Is it possible to solve it there instead?

Issue #507, opened 2023-08-08
Forwarded messages from mautrix-telegram does not render


q234rty:
### Bug Description

Forwarded messages on the telegram side bridged by mautrix-telegram does not render.

### Steps to Reproduce

1. Send `<tg-forward><blockquote>message</blockquote></tg-forward>` in element desktop with /html

2. Observe that message does not render.

### Expected Behavior

message is rendered fine.

### App Version

rc1.13.0-2

### Additional Platform Information

Arch Linux

### Additional Context

I _think_ this is a regression from https://github.com/krille-chan/fluffychat/commit/796a6fdae181eb1118bffdb05f2e6e3d378fe6c4

krille-chan:
Can you please test this with the newest main version? I've added it to the allowed tags

q234rty:
I tried compiling from git but it still doesn't work for some reason

jahway603:
I'm running the latest version on F-Droid, Version 1.16.0 (3523), Added on 2023-12-10, and this issue is still happening.

krille-chan:
Fixed in latest commit

PR #574, opened 2023-09-16
fix: tg-forward

Decodetalkers:
replace tg-forward block to ''

this make tg-forward can be shown

Log:

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -34,15 +34,34 @@ class HtmlMessage extends StatelessWidget {
     // We do *not* do this in an AST and just with simple regex here, as riot-web tends to create
     // miss-matching tags, and this way we actually correctly identify what we want to strip and, well,
     // strip it.
-    final renderHtml = html.replaceAll(
-      RegExp(
-        '<mx-reply>.*</mx-reply>',
-        caseSensitive: false,
-        multiLine: false,
-        dotAll: true,
-      ),
-      '',
-    );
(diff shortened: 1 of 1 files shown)


Decodetalkers:
@krille-chan I have fixed tg-forward, make the part visible, can you review it?

krille-chan:
Can you elaborate more what this is about?

Decodetalkers:
telegram has a message is <tg-forward>, the message my side received is like 
```
<tg-forward>
    <blockquote>
message content
   </blockquote>
</tg-forward>
```
because the xml has `<tg-forward>` part, the part of <blockquote> can not be rendered, it become empty, then I cannot view the forward message here

q234rty:
I guess this fixed #507 ?

lib/pages/chat/events/html_message.dart
```diff
@@ -34,15 +34,34 @@ class HtmlMessage extends StatelessWidget {
...
+        .replaceAll(
+          RegExp(
+            '<mx-reply>.*</mx-reply>',
+            caseSensitive: false,
+            multiLine: false,
+            dotAll: true,
+          ),
+          '',
+        )
+        .replaceAll(
```
krille-chan, line 47: I don't think we need two replaceAll methods here. With regex it should be possible to do it in one by selecting the content inside of tg-forward.

krille-chan:
Instead of doing an expensive `.replaceAll()` action on all messages, I would prefer to just add `tg-forward` as another allowed tag and render it the same way as blockquotes

PR #593, opened 2023-09-26
Update styles.xml

HotarunIchijou:


krille-chan:
Sorry for me this does not work. Can you add more information about what this does and why it is only added to night mode?

HotarunIchijou:
I tried to apply [edge to edge](https://developer.android.com/develop/ui/views/layout/edge-to-edge) but I'm not really familiar with flutter. I guess it should be applied this way:
```class MyApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return SystemChrome(
      systemUiOverlayStyle: SystemUiOverlayStyle(
        statusBarColor: Colors.transparent,
      ),
      child: Scaffold(
        // ...
      ),
    );
  }
}
```

This code was generated by Google Bard, so I'm not really sure if it will work

HotarunIchijou:
Also the reason why I tried to apply it only for dark theme is because it's more noticeable

PR #643, opened 2023-11-03
fix: Properly capitalize Linux window title

kra-mo:


krille-chan:
Thank you very much :)

PR #646, opened 2023-11-04
chore: added android audio sharing intent

### LINKED ISSUE #429: Fluffychat missing from audio share intents
#### Description

I may want to share an audio file from a different app, but fluffychat does not appear as a share intent for any audio file.

I must instead locate the audio in my filesystem or share to schildichat instead.

#### To Reproduce

Share an audio file from a voice recorder or music app. Try to find fluffychat in the list.

#### Additional information:
- Device: Galaxy S20
- OS and OS version: Android 12
- Installed version of FluffyChat: 1.6.2
- (Android only) Which store are you using: Gitlab Release
- (Android only) Are Google Services available: Yes, but I use UP

aryanA101a:
fix #429

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -95,6 +95,11 @@
                 <category android:name="android.intent.category.DEFAULT" />
                 <data android:mimeType="video/*" />
             </intent-filter>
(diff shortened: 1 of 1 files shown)


krille-chan:
Thank you very much <3

PR #651, opened 2023-11-06
Fix Dockerfile: install jq in the builder image

douardda:
It seems jq is now required by the prepare-web.sh script, so add it in the builder image.


krille-chan (approved):
Thank you so much 👌

Issue #500, opened 2023-08-06
No way to accept an in coming voice call


K4LCIFER:
### Bug Description

When someone calls you, all you see is a message that gets sent saying "[Username] started a call", but you have no method of accepting it. 

### Steps to Reproduce

1. Receive a voice call
2. observe that you have no means of accepting it

### Expected Behavior

I expect to have some buttons to interact with the call e.g. accept, decline, etc.

### App Version

1.13.0

### Additional Platform Information

Google Pixel 6, Android 14 beta

PR #652, opened 2023-11-06
fix: call gui not visible on devices with android version > android 11

### LINKED ISSUE #499: Starting a voice call from Fluffychat Android doesn't open any GUI for the call
### Bug Description

When you start the call from the app, you have no way of interracting with it. All it does is sends a message stating "[Username] started a call". The only way to end the call is for the other person to hang up, and you close the app.

### Steps to Reproduce

1. Go to a chat with someone
2. Start a call 
3. observe that no GUI pops up

### Expected Behavior

I would expect some form of GUI to pop up to allow me to have basic interraction with the call

### App Version

1.13.0

Google Pixel 6

### Additional Context

This may be a feature request instead, but I'm assuming this functionality already exists, and isn't being triggered.

aryanA101a:
fixes #499 #500 

On devices greater than android 11, `read_phone_state` permission is required to be consented during runtime.

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -1190,6 +1190,11 @@
         "type": "text",
         "placeholders": {}
     },
+    "phonePermissionDeniedNotice": "Phone permission denied. Please grant it to be able to properly send and recieve calls.",
+    "@phonePermissionDeniedNotice": {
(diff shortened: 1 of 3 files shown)


Integral-Tech:
@aryanA101a Why this pull request was closed?

PR #659, opened 2023-11-09
Change dead link in README

RasmusRendal:
Seems like it was removed from the MSC repo at some point.


krille-chan (approved):
Thank you so much 👌

PR #674, opened 2023-11-13
fix: call gui not visible on devices with android version > android 11

### LINKED ISSUE #499: Starting a voice call from Fluffychat Android doesn't open any GUI for the call

aryanA101a:
fixes #499 #500

### DIFF
--- a/lib/utils/voip/callkeep_manager.dart
+++ b/lib/utils/voip/callkeep_manager.dart
@@ -1,13 +1,16 @@
 import 'dart:async';
+import 'dart:developer';
 
 import 'package:flutter/material.dart';
 
 import 'package:callkeep/callkeep.dart';
+import 'package:device_info_plus/device_info_plus.dart';
 import 'package:flutter_foreground_task/flutter_foreground_task.dart';
 import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:matrix/matrix.dart';
 import 'package:permission_handler/permission_handler.dart';
 
+import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:fluffychat/utils/voip_plugin.dart';
 
 class CallKeeper {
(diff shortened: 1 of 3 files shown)


aryanA101a:
@krille-chan please have a look at it

lib/utils/voip/callkeep_manager.dart
```diff
@@ -365,3 +385,8 @@ class CallKeepManager {
     setCallHeld(event.callUUID, event.hold);
   }
 }
+
+Future<PermissionStatus?> tryReadPhoneStatePermission() async {
+  final info = await DeviceInfoPlugin().androidInfo;
```
krille-chan, line 399: Doesn't this need a platform check?

lib/utils/voip/callkeep_manager.dart
```diff
@@ -167,7 +168,26 @@ class CallKeepManager {
     Logs().v('[onPushKitToken] token => ${event.token}');
   }
 
-  Future<void> initialize() async {
+  Future<void> initialize({BuildContext? context}) async {
+    tryReadPhoneStatePermission().then((pStatus) {
+      if (context != null) {
+        if (pStatus == PermissionStatus.denied ||
+            pStatus == PermissionStatus.permanentlyDenied) {
+          ScaffoldMessenger.of(context).showSnackBar(
```
krille-chan: What happens if a user will permanently deny the permission? Will he gets a snackbar on any app start with the same content?

aryanA101a: Okay, I'll add a different message to the snackbar for permanently denied.
What do you prefer more? After getting permanently denied, show the snack bar only once or everytime the user opens the app?

lib/utils/voip/callkeep_manager.dart
aryanA101a: Yes, I'll fix it

aryanA101a:
@krille-chan I have made the changes, please have a look at it.

mayemu:
This is still an issue, should be opened

PR #704, opened 2023-11-30
fix: properly scale emote-only messages with custom emotes

TheOneWithTheBraid:
image

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -1,4 +1,4 @@
-import 'package:flutter/material.dart';
+import 'package:flutter/material.dart' hide Element;
 
 import 'package:collection/collection.dart';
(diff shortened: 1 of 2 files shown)


TheOneWithTheBraid:
Rebased onto current changes to HTML renderer code, I'd appreciate some review, @krille-chan !

lib/pages/chat/events/html_message.dart
```diff
@@ -525,3 +553,15 @@ class MatrixPill extends StatelessWidget {
     );
   }
 }
+
+FontSize? tryGetParentFontSize(ExtensionContext context) {
```
TheOneWithTheBraid, line 557: This function is basically the relevant part scaling custom emotes into the right parenting text size. I'm sure it could be better or "more complete" but from everything I see as of now how we use font sizes in the HTML renderer, it should work in any occurring case.

PR #764, opened 2024-01-02
GHSA-pcvr-f2f2-xq7m: add Desktop database encryption

TheOneWithTheBraid:
- refactor Database factory code
- use cipher for Desktop FFI as well
- build SQfLite factory with sqlcipher support


lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
```diff
@@ -52,63 +51,61 @@ Future<MatrixSdkDatabase> _constructDatabase(Client client) async {
...
+  } else {
+    final path = await getApplicationSupportDirectory();
+    final sqlFilePath = '$path/${client.clientName}.sqlite';
 
-  final path = await getDatabasesPath();
-  const passwordStorageKey = 'database_password';
-  String? password;
+    // migrating from petty unreliable `getDatabasePath`
+    // See : https://pub.dev/packages/sqflite_common_ffi#limitations
+    await _migrateLegacyLocation(sqlFilePath, client.clientName);
```
krille-chan: Please do only one thing per Pull Request!

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
```diff
@@ -52,63 +51,61 @@ Future<MatrixSdkDatabase> _constructDatabase(Client client) async {
...
+    fileStoragePath: fileStoragePath,
     deleteFilesAfterDuration: const Duration(days: 30),
   );
 }
+
+Future<void> _migrateLegacyLocation(
+  String sqlFilePath,
+  String clientName,
+) async {
+  final oldPath = getDatabasesPath();
```
krille-chan: This does not work as it returns a future

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
```diff
@@ -52,63 +51,61 @@ Future<MatrixSdkDatabase> _constructDatabase(Client client) async {
...
+  final maybeOldFIle = File('$oldPath/$clientName.sqlite');
```
krille-chan: Typo in "File"

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
```diff
@@ -52,63 +51,61 @@ Future<MatrixSdkDatabase> _constructDatabase(Client client) async {
...
+  if (await maybeOldFIle.exists()) {
```
krille-chan: Should check before if the two path strings are already equal. Otherwise it would copy the database every time the app opens

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
```diff
@@ -52,63 +51,61 @@ Future<MatrixSdkDatabase> _constructDatabase(Client client) async {
...
+    await maybeOldFIle.copy(sqlFilePath);
+    await maybeOldFIle.delete();
+  }
+}
+
+Future<void> _setPragmaKey(Database db, String cipher) =>
```
krille-chan: This function name implies that this can set any pragma key but it can only set one. Suggestion for a better name: `_setCipherPragmaKey()`

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/sqlite_ffi/io.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'dart:math';
+
+import 'package:sqlite3/open.dart';
+
+void ffiInit() {
+  open.overrideFor(OperatingSystem.linux, sqlcipherOpen);
+}
+
+DynamicLibrary sqlcipherOpen() {
+  // Taken from https://github.com/simolus3/sqlite3.dart/blob/e66702c5bec7faec2bf71d374c008d5273ef2b3b/sqlite3/lib/src/load_library.dart#L24
```
krille-chan: This needs much much more information why this is needed. The link only shows me that the exact same code is already in a different library? Why do we need it here then and can't just use what is there? This must be explained here in more detail

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
TheOneWithTheBraid: Named it `_applySQLCipher`, I think that's pretty much what it does, right ?

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/sqlite_ffi/io.dart
TheOneWithTheBraid: Added the corresponding doc link in he file header.

lib/utils/matrix_sdk_extensions/flutter_matrix_dart_sdk_database/builder.dart
TheOneWithTheBraid: True, removed. That's gonna be another PR. Initially considered is close enough since it's both related to issues with the DB migration but of course you are right here.

TheOneWithTheBraid:
Thanks for the review, @krille-chan . Fixed the issues you mentioned.

krille-chan (changes requested):
What happens with existing Linux users? Do they all get logged out with this when the database now tries to encrypt everything but the existing database is not using this pragma?

TheOneWithTheBraid:
> What happens with existing Linux users? Do they all get logged out with this when the database now tries to encrypt everything but the existing database is not using this pragma?

I just wrote a migration for our proprietary matrix client, I'm pretty sure I can share it later - also for work related implementations if you are interested.

It currently does a byte check on whether the database file is plain or encrypted, in case it is plain, it starts an SQLCipher migration into an (encrypted) new file and replaces the plain DB file with the new one in case the migration was successful.

TheOneWithTheBraid:
I revamped this PR into a general migration from separated SQLite implementations to one using FFI (which _does_ support encryption). I hence included the location migration too - since unencrypted DBs are migrated too, I'd feel weird about not migrating the location at the same time as well.

Draft: Uses override of SDK.

krille-chan:
I see this error on every app start now:

```sh
*** sqflite warning ***

You are changing sqflite default factory.
Be aware of the potential side effects. Any library using sqflite
will have this factory as the default for all operations.

*** sqflite warning ***
```

krille-chan:
on macOS this works fine. on iOS (in simulator) I see this error in the logs:

```sh
flutter: error Bad state: SQLCipher library is not available, please check your dependencies! during open, closing...
```

TheOneWithTheBraid:
We need to add to the wiki how to install SQLCipher I guess :

> macOS and iOS: Depends on the SQLCipher pod. IMPORTANT NOTE: Bad things will happen if you depend on any other package linking the regular sqlite3 library. Please be sure to read the advisory before using this package.

But that's a task for the moment the PR is merged.

TheOneWithTheBraid:
> I see this error on every app start now:
> 
> ```shell
> *** sqflite warning ***
> 
> You are changing sqflite default factory.
> Be aware of the potential side effects. Any library using sqflite
> will have this factory as the default for all operations.
> 
> *** sqflite warning ***
> ```

That's not an issue - we enforce sqlite to use the SQLCipher dynamic library - if we should ever load both, we'd run into issues, we hence enforce sqlcipher for all calls of it.

krille-chan:
> > macOS and iOS: Depends on the SQLCipher pod. IMPORTANT NOTE: Bad things will happen if you depend on any other package linking the regular sqlite3 library. Please be sure to read the advisory before using this package.

Does not work. I added `pod 'SQLCipher', '~>4.0'` to the podfile and sqlcipher is also present in Podfile.lock but still get this error

krille-chan:
> > I see this error on every app start now:
> > ```shell
> > *** sqflite warning ***
> > 
> > You are changing sqflite default factory.
> > Be aware of the potential side effects. Any library using sqflite
> > will have this factory as the default for all operations.
> > 
> > *** sqflite warning ***
> > ```
> 
> That's not an issue - we enforce sqlite to use the SQLCipher dynamic library - if we should ever load both, we'd run into issues, we hence enforce sqlcipher for all calls of it.

We won't have if we just use the defaults, means sqflite_sqlcipher package, right? I see no reason why not using this package for Android, iOS and macOS as it seems to be widely used everywhere and our own solution only for Linux and Windows

krille-chan:
What about you contribute support for Linux and Windows to the sqflite_sqlcipher package instead?

TheOneWithTheBraid:
> > > I see this error on every app start now:
> > > ```shell
> > > *** sqflite warning ***
> > > 
> > > You are changing sqflite default factory.
> > > Be aware of the potential side effects. Any library using sqflite
> > > will have this factory as the default for all operations.
> > > 
> > > *** sqflite warning ***
> > > ```
> > 
> > 
> > That's not an issue - we enforce sqlite to use the SQLCipher dynamic library - if we should ever load both, we'd run into issues, we hence enforce sqlcipher for all calls of it.
> 
> We won't have if we just use the defaults, means sqflite_sqlcipher package, right? I see no reason why not using this package for Android, iOS and macOS as it seems to be widely used everywhere and our own solution only for Linux and Windows

Nope, if we'd use sqlite_sqlcipher and any other package loads a regular SQLite dynlib, we'd run into exactly the same trouble.

TheOneWithTheBraid:
> What about you contribute support for Linux and Windows to the sqflite_sqlcipher package instead?

Hmmm, I actually highly prefer the FFI approach : sqflite has one more layer : It's using the Flutter plugin API to speak with a native implementation, the native implementation then speaks with the sqlite / sqlcipher library. When using FFI, we reduce this to direct communication to the dynlib on all platforms which shrinks the app's native code. Since the FFI approach supports all platforms (unlike sqflite_sqlcipher), I think that's the way to go - I'd rather invest some time into further debugging the iOS setup, it might be simply conflicting with some other package loading SQL related stuff. I can take care of this once I'm home with access to mac hardware again.

TheOneWithTheBraid:
I cannot reproduce the build issue on my MacBook. Did you install sqlcipher ?

```
brew install sqlcipher
```

That's everything I needed. Still, I added a small patch to the Podfile checking for other packages shipping sqlite, I will push it in a moment.

kimimaru4000:
Hi @TheOneWithTheBraid @krille-chan, what's the blocker on getting this merged?

TheOneWithTheBraid:
> Hi @TheOneWithTheBraid @krille-chan, what's the blocker on getting this merged?

I just rebased and could only kindly ask @krille-chan to have a look at it once again.

krille-chan:
> > Hi @TheOneWithTheBraid @krille-chan, what's the blocker on getting this merged?
> 
> I just rebased and could only kindly ask @krille-chan to have a look at it once again.

Well I still do not agree that we should introduce such a complexity. Especially for such a critical component as the database is. We got burned too often by lost sessions, that I only want to use the most canonical direct obvious way for databases and that's using sql_cipher on Android and iOS.
However, sql_cipher already supports macOS.

jane400:
Hi, I'm a user of Fluffychat and noticed that my datebase "got optimized" with a recent upgrade (the update was delayed in Alpine Linux). But it's not encrypted now and can be read out by every application that has access to the sqlite database. Fluffychat should encrypt it's database on every platform, not just mobile ones. As I see that this MR accomplishes this it should IMO merged immediately as a security fix.


> Well I still do not agree that we should introduce such a complexity. Especially for such a critical component as the database is.
I don't understand the code perfectly yet, as I'm not used to develop in dart, but it appears to be a refactor of `lib/utils/matrix_sdk_extensions/flutter_matrix_sdk_database_builder.dart` into multiple files and uses the same previous code for mobile now on desktop. It also moves handling of the encryption into the `SQfLiteEncryptionHelper` into the matrix-dart-sdk.  I don't see added complexity here, just moved code.  As I understand multiple companies use the matrix-dart-sdk in their proprietary matrix client, I assume they use this API too, so it should be well tested. No need to keep handling it Fluffychat.

TheOneWithTheBraid:
@krille-chan I can only agree to what's just been said. This MR simplifies the code instead of adding further complexity - the implementation we have in the SDK works like a charm, at least in our proprietary products.

I consider it as a way more simple implementation to have 100 % of the same code for Desktops as for Mobile. I do not think that `sqflite_sqlcipher` has any advantages over `sqlcipher_flutter_libs`. Both are pretty well maintained and broadly used - with the difference that the FFI implementation is cleaner (as in one less layer in between) and works on all platforms.

krille-chan:
It definitely adds complexity, as it is with this changes using a bunch of code which is copy-pasted from another library: https://github.com/famedly/matrix-dart-sdk/blob/main/lib/src/database/sqflite_encryption_helper/io.dart instead of abstracted. What happens in the long term when this breaks and needs to be fixed? We cannot just rely on the upstream library then but need to fix it by ourself. This changes from a simple `openDatabase(password: ...)` to a lot of new code which I have to maintain then. What if a bug appeared in it or a security leak? I don't want to have this responsibility as I got burned too often for users losing their session.

> But it's not encrypted now and can be read out by every application that has access to the sqlite database.

FluffyChat has **not** encrypted the database on Linux in the versions before as well. The hive database was unencrypted and could be read by anyone. You should not rely on that applications encrypt their local cache but rely on sandboxing. Afaik on Flatpak the Sandbox is basically a lie (https://flatkill.org/) so I recommend using the snap version.

TheOneWithTheBraid:
Hmmm, but that's exactly why the code is abstracted in the Matrix Dart SDK, isn't it ? This exactly prevents from some update breaking it - since the database encryption related code is exactly _not_ part of FluffyChat but rather part of the SDK. I don't get that objection. And also, instead of two SQLite related implementations, it's now reduced to a single one - since the code from the SDK supports *any* SQLite-enabled platform instead of just mobile ones.

I don't think anyone was talking about Flatpak here. Single package formats should not be the reason to store something more or less secure, though I fully agree on the disadvantages about Flatpak.

kimimaru4000:
What information does Fluffychat's database store? The `PRIVACY.md` contains [this section](https://github.com/krille-chan/fluffychat/blob/main/PRIVACY.md#database) which just links to the Hive package.

TheOneWithTheBraid:
This is a workaround for particular Linuxes. If a distribution decides to ship a patched version, it will be used. I tbh don't think it is necessary - but not evil either. It simply makes packaging easier in some cases - and if a distributor does not ship this bundle, we're always at the regular SQLCipher.

krille-chan:
I would be fine if we say we keep only change the database for desktop and keep the iOS and android database as it is for now. Would prefer to change one thing at a time anyway. Maybe in the future the situation is different but for now I would suggest we only change how the desktop database is handled. Would this bis a compromise @TheOneWithTheBraid ?

PR #768, opened 2024-01-03
chore: migrate sqlite file location

TheOneWithTheBraid:
- avoid using getDatabasePath
- use getApplicationSupportDirectory instead

Depends : #764 (rebased)


krille-chan:
Has been implemented in the other PR already

PR #777, opened 2024-01-07
build: make dependabot also monitor github-actions

Mikaela:


krille-chan (approved):
Ah good idea. Thanks :)

PR #798, opened 2024-01-13
fix: HTML renderer edge-case without top-level element

TheOneWithTheBraid:
The PR contains two commit : 

- first one only fixes the relevant issue
- second removes previous workaround applying <p> elements around links since that's less efficient and no longer needed with the patch also catching several top-level elements on message base

In case you only want to use the pure workaround without the cleanup, feel free to cherry-pick the first commit only.

Test on message content :

```
<p> Test paragraph !</p>
<pre>
Some
Content
Here
</pre>
```

Note that the sample has no containing HTML parent but consists of _several_ Nodes instead of only one.

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -88,7 +88,13 @@ class HtmlMessage extends StatelessWidget {
       padding: HtmlPaddings.only(left: 6, bottom: 0),
     );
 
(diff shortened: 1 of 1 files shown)


krille-chan:
Can you please explain what this actually does? I find it super confusing. The commit message says it fixed the bug with the case for two top level elements, while this is already fixed? Now this changes the whole code but why? What does it make better? What is the purpose of this?

PR #846, opened 2024-01-27
Added Docker deploy step to release

jgkawell:
This adds a new job to the release pipeline that builds and publishes the Docker image to the GitHub container registry. Example of it running in my fork: https://github.com/jgkawell/fluffychat/pkgs/container/fluffychat

This will help all of us who run the web version in Docker so that we can simply update the tag on our deployment manifest and grab the latest.


krille-chan (approved):
Thanks, we can try it out and see how it works :)

PR #855, opened 2024-01-31
feat: calls

td-famedly:
:D


lib/config/app_config.dart
```diff
@@ -53,7 +53,7 @@ abstract class AppConfig {
   static bool livekitEnabledCalls = true;
   // static const livekitServiceUrl = 'https://famedly-livekit-server.teedee.dev';
   static const livekitServiceUrl = 'https://livekit-jwt.call.element.dev';
```
blaggacao: Out of couriosity: are these going to be hosted services for the app? Or are users expected to run livekit alongside their own homeservers?

And maybe one side-question, if allowed: is there an interoperability with, say, call.element.io so that a group call can be shared with outside users (non-fluffy/non-matrix)?

All-in-all though, this looks pretty exciting. Thank you for working on this!

alexander-potemkin: I'm not an author, but from my humble knowledge and understanding - it shall work both ways, as livekit seems to be a protocol.

The only question / problem is that Element doesn't currently support custom livekit server (https://github.com/element-hq/element-meta/issues/2371).

I hope the server string will be configurable with FluffyChat, livekit is not that much complicated to be self-hosted.

alexander-potemkin:
@krille-chan , would you be open to some help in that? Or calls are outside of FluffyChat roadmap?

krille-chan:
> @krille-chan , would you be open to some help in that? Or calls are outside of FluffyChat roadmap?

Currently that's a project of @td-famedly 😊 you can contact him for collaboration maybe?

lib/config/app_config.dart
td-famedly: > if allowed: is there an interoperability with, say, call.element.io so that a group call can be shared with outside users (non-fluffy/non-matrix)?

there's some differences in the key sharing right now (to-device vs room events specifically), but once that gets solved (the fosdem demo f.ex had a patch to use room events for key sharing and be compatible with element), fluffychat should just be able to use call.element.io or any other livekit SFU for calls. (The current logic is look at older members in the call and try to use their SFU, because cross SFU communication is not a thing yet)

> I hope the server string will be configurable with FluffyChat, livekit is not that much complicated to be self-hosted.

That sounds doable. @krille-chan I assume we can add a developer setting (maybe hidden) to change the default livekit url?  Again this URL only matters if you are the one starting the call. Both element call and fluffychat should just use the SFU from any existing participants state event if there is already a call going on. 

tldr: which SFU to choose should not be a problem, key sharing implementation differences are a bigger blocker atm

krille-chan: Sure we can. While I still wonder what makes more sense: Set the livekit server in the well-known for a server wide config or set it in the app as an app config?

alexander-potemkin:
Apologies for the delay in getting back to you!

@krille-chan , thank you!
@td-famedly , what is the best way to contact you? Or shall we keep discussion here? What kind of help could be of use? I'm not quite fast on that, unfortunately, but I shall be able to be of some help.

predictron-cloud:
Hi guys, This implementation looks amazing. It is likely to provide great user experience as it uses livekit and flutter_callkit_incoming. When it could be merged and released? Do you need any help with it?

sadorowo:
Hi, calls are the only feature stopping me from switching to FluffyChat. In my opinion Element is slow and I would like to switch to the mature alternative. I appreciate all the work that you guys have put in this app. This PR would resolve issues with calls, and as I can see, we have Livekit implementation here, which is the latest Matrix call standard. Unfortunately this PR is a bit abandoned. When it could be merged?

lib/config/app_config.dart
sadorowo: > Sure we can. While I still wonder what makes more sense: Set the livekit server in the well-known for a server wide config or set it in the app as an app config?

We already have this field dedicated for supplying the Livekit server url:

`GET /.well-known/matrix/client`
```json
"org.matrix.msc4143.rtc_foci": [
    {
        "type": "livekit",
        "livekit_service_url": "https://livekit.example.com"
    }
]
```

Note that we have to search for url based on a `"type"` key, not indexes, on some advanced setups there might be some other values which are not relevant to call feature

alexander-potemkin: App level config, as I see, will make it complicated to use anything but centralized installation,

klarlichter:
We create a fix and would push the fix that call feature is work. Should we do this ? :)

alexander-potemkin:
> We create a fix and would push the fix that call feature is work. Should we do this ? :)

Please, don’t hesitate to do that :)

sadorowo:
As I can see this PR was a part of the Google Summer of Code 2022. You can read more about it [here](https://technodisaster.medium.com/gsoc22-at-matrix-org-91919707742e)

klarlichter:
That fix are here 

https://github.com/krille-chan/fluffychat/pull/2140

Hope @krille-chan  can merge it into main that all can use now call and video function.

krille-chan:
> That fix are here
> 
> #2140
> 
> Hope @krille-chan can merge it into main that all can use now call and video function.

Thank you so much. Maybe @td-famedly you can take a look on this?

standortdigital:
> > That fix are here
> > #2140
> > Hope @krille-chan can merge it into main that all can use now call and video function.
> 
> Thank you so much. Maybe @td-famedly you can take a look on this?

We was successfully test our fix inside our own build. It is working like expected. I was create a merge request on main branch.

jackedproxy:
Any update on this one?
Been a while since the last update.

czadikem:
Is this ever going to come to fruition?

onestacked:
I asked @td-famedly in the Webrtc room:

> and don't plan to anytime soon, so at some point in the future probably, but tbh not anytime soon

https://matrix.to/#/!DutPDfXYX_Hi3CekWhkPVrG6WrZI73fWQUybhjPdMAk/$P_nT-WgcHHWP17oWx0aEyFu2wlJU90ITGnM1No_VhvM?via=element.io&via=matrix.org&via=codestorm.net

There is also an [issue](https://github.com/livekit/client-sdk-flutter/issues/974) with the Livekit Dart SDK that would limit this to unencrypted room and shared keys (like in SPA mode), but the limited support should not be to much work since the dart SDK has support for that already. Commet already [supports unencrypted calls](https://github.com/commetchat/commet/pull/476) using this.

Issue #334, opened 2023-07-15
FluffyChat doesn't expose language option in systen settings


krille-chan:
#### Description

On iOS, the app language can generally be specified as other than the primary system language in Settings, however FluffyChat doesn't allow this.

Previously reported in https://gitlab.com/famedly/fluffychat/-/issues/746#note_798086929.

#### To Reproduce

1. Open the Settings app
2. Scroll to FluffyChat
3. See that there is no language option.

If you check other apps, a lot have a language option there, for example 1.1.1.1, Adobe Acrobat, AdGuard, Authy, Bitwarden, Brave, and conmsidering I am still listing letter A of apps I consider popular, it's very common and I consider it an issue with FluffyChat.

#### Additional information:

- Device: iPhone SE2020
- OS and OS version: iOS
- Installed version of FluffyChat: 
- (Android only) Which store are you using: Testflight

Mikaela:
I retitled this as I think I was the original opened at GitLab and the issue also affects Android.

Going to Settings -> App Info -> FluffyChat should expose a language setting, but it's not there.

If I look at other apps such as YouTube Music or Headspace, there is a language menu.

Mikaela:
Implementation in Android app Voice Notify in case it's helpful:

- https://github.com/pilot51/voicenotify/commit/65af1cd7559e9326a2fc7fd5bb28d598600873ad

Mikaela:
This seems to be the Android documentation:

- https://developer.android.com/guide/topics/resources/app-languages

Mikaela:
- https://developer.apple.com/news/?id=u2cfuj88

looks like iOS equivalent

Mikaela:
Blocked by https://github.com/flutter/flutter/issues/109842?

Mikaela:
Upstream

Mikaela:
I just marked https://github.com/krille-chan/fluffychat/issues/2147 as a duplicate of this issue four days ago.

PR #888, opened 2024-02-11
android/app/build.gradle: attempt to enable language settings

Mikaela:
Ref: #334

I have no idea if this is the correct file or way to do it.


Mikaela:
This is beyond my abilities and seems like an upstream issue. https://github.com/krille-chan/fluffychat/issues/334#issuecomment-1937578376

PR #889, opened 2024-02-11
Fix swapped icons for mark read/unread and pin/unpin

Bubu:
If any selected room is not marked unread
(anySelectedRoomNotMarkedUnread), then show the 'mark unread' action
icon, else show the 'mark as read' icon.

This makes it consistent with the notification on/off action.

Similarly to the read/unread toggle this was also using the wrong icon.

If any selected room is not favourite, then show the pin icon (which
should represent the "mark as favourite" action as it corresponds to the
filled pin shown next to a favourited room. Otherwise show the outline
icon, representing the "unpin" action.


krille-chan (approved):
Thank you very much :)

PR #890, opened 2024-02-11
fix: properly initialize hideUnimportantStateEvents setting

Bubu:
This setting was not applied on app restart but always initialized to the default value. The only way change this was to go into settings and change the toggle twice after an app restart.


krille-chan (approved):
Thank you very much for the fix 👍

PR #892, opened 2024-02-11
feat: make showing user presence info optional

Bubu:
895de76e70466616d1baa45f21fa0a3fdeb30ec7 replaced the stories feature with presence status messages. Stories were an optional feature but right now presence information can only be disabled on a homeserver level.

Introduce a setting to make this feature optional on a client level.

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -1734,6 +1734,16 @@
         "type": "text",
         "placeholders": {}
     },
+    "presenceStyle": "Presence:",
+    "@presenceStyle": {
+        "type": "text",
+        "placeholders": {}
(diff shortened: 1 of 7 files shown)


Bubu:
I would have liked to have a popup/context menu on long-press/right-click that allows hiding this section [1] but right now this is beyond my flutter skills. I.e. specifically I can't figure out how to rebuild the stateless ChatListViewBody or StatusMessageList widgets when an AppConfig setting changes.


[1] This is what telegram does as well now, after they introduced stories

krille-chan (approved):
LGTM 👍 and thanks for the info how Telegram handles this. Sounds reasonable then to me

PR #906, opened 2024-02-25
ci: add a license compliance check

selfisekai:
this is to prevent a non-free license from appearing in the project, as previously happened with dart_code_metrics, which, unnoticed, re-licensed from MIT to BUSL-1.1. this only got noticed because they've then switched entirely to a paid service-as-a-software-substitute model, and the unmaintained module stopped working with dart 3.3/flutter 3.19.

caveats:
- sky_engine (aka dart:ui) is licensed BSD-3-Clause itself, but includes this license file with whatever their C/C++ dependencies put in the repo, automatically amended by a bot the whole time. https://github.com/flutter/engine/blob/main/sky/packages/sky_engine/LICENSE
- dependencies can declare their license incorrectly, or miss something. like the sqlite3 module doesn't mention blessing. but this check is still a step forward


krille-chan:
Ah thank you very much. That looks like a great improvement

krille-chan (approved):

PR #919, opened 2024-03-02
Add settings to keep images aspect ratio

BhasherBEL:
This PR add a settings to force images to keep their initial aspect ratio in chat.

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -154,6 +154,11 @@
         "type": "text",
         "placeholder": {}
     },
+    "keepImageAspectRatio": "Keep image aspect ratio",
+    "@keepImageAspectRatio": {
+        "type": "text",
+        "placeholder": {}
+    },
     "badServerLoginTypesException": "The homeserver supports the login types:\n{serverVersions}\nBut this app supports only:\n{supportedVersions}",
     "@badServerLoginTypesException": {
         "type": "text",
--- a/lib/config/app_config.dart
+++ b/lib/config/app_config.dart
@@ -46,6 +46,7 @@ abstract class AppConfig {
   static bool hideUnimportantStateEvents = true;
   static bool separateChatTypes = false;
   static bool autoplayImages = true;
+  static bool keepImageAspectRatio = false;
   static bool sendTypingNotifications = true;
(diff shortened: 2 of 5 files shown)


M00NJ:
I think it makes sense to retain the aspect ratio by default. When sending a screenshot often times important information is not visible in the chat and I worry the other person might not bother the open the image (since they don't know that I highlighted something at the top or bottom for example). This is also how all the messanger apps I've used so far handle it.

BhasherBEL:
I agree that imo this should be the default behavior, as in many apps.

For now, the feature looks a bit ugly, as it simply force the picture to fit inside the box. It could instead resize the outside borders to also match the picture.

image

If you agree for both, I will implement them.

krille-chan:
If you find a solution which looks good for all possible aspect ratios of pictures, sure. The current size should then just be a max width/height

BhasherBEL:
The new commit should change what we have discussed. The size are constraints between 100-400 for width and 100-500 for height.

Here is how it looks like now:
image

Here is how it handle extreme cases. 
image

krille-chan:
Doesn't seem to work yet. When I send a photo which is very wide or very high but also very thin it looks like this:

Bildschirmfoto vom 2024-03-10 15-04-52

BhasherBEL:
Did you use the last commit ? It's looks like the previous behavior where the box didn't scale. What happens if you use a standard picture (i.e. not extreme) ?

krille-chan:
> Did you use the last commit ? It's looks like the previous behavior where the box didn't scale. What happens if you use a standard picture (i.e. not extreme) ?

Yes this screenshot is from the newest version. The aspect ratio needs to work with all possible ratios, even with extreme ones

BhasherBEL:
> > Did you use the last commit ? It's looks like the previous behavior where the box didn't scale. What happens if you use a standard picture (i.e. not extreme) ?
> 
> Yes this screenshot is from the newest version. The aspect ratio needs to work with all possible ratios, even with extreme ones

Definitely agree. But it's quite strange that it's still squared on your screenshot. It should never be squared again, except for *real* squared pictures.

krille-chan:
Has now been implemented with a different approach. Anyway, thank you very much for the attempt

PR #920, opened 2024-03-02
Fix send on enter

BhasherBEL:
This PR aims to fix the fact that send on enter is not working as expected at least on desktop. The issue seems to come from the fact that the settings is null by default, but turned on in the settings.

By providing a default value, this issue should be prevented.

### DIFF
--- a/lib/config/app_config.dart
+++ b/lib/config/app_config.dart
@@ -48,7 +48,7 @@ abstract class AppConfig {
   static bool autoplayImages = true;
   static bool sendTypingNotifications = true;
   static bool sendPublicReadReceipts = true;
-  static bool? sendOnEnter;
+  static bool sendOnEnter = false;
   static bool showPresences = true;
   static bool experimentalVoip = false;
   static const bool hideTypingUsernames = false;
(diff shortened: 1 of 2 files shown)


krille-chan:
Sorry, this does not work. There are three states of the config "sendOnEnter":
- true: It should aways send on enter (purpose is to make it possible on mobile as well, so iPad users with keyboard are happy)
- false: It should never send on enter (purpose is to make it possible on desktop/web as well, so mobile web users and linux phone users are happy)
- null: It fallbacks to the platform, so acts like false on Android&iOS and like true on web&Desktop

We need all three states here to fulfill this

krille-chan:
> The null value make it strange, at least for desktop. The displayed state in settings doesn't match the actual behavior.

It doesn't make it strange at all. `null` means here `default` and the default depends on the platform

nc7s:
Sorry to dig this out but this is not how toggles work. A toggle has, and only has, 2 states: on and off. Period. There is no 3rd state. Unless it's a 3-way toggle. So there shouldn't be a 3rd state even if it's hidden in the code. This is basic UI/UX: state/code/logic should be consistent with interface.

IMHO the correct thing to do about "platform defaults", is to set the state to the platform default on first start, and leave it to the user ever after.

PR #924, opened 2024-03-04
Switch nginx to one that doesn't need root

halkeye:


halkeye:
Confirmed:

deployed @ https://fluffychat.g4v.dev/
config: https://github.com/halkeye/home-k8s/blob/main/charts/fluffychat/values.yaml#L7

BhasherBEL (approved):

krille-chan:
I don't know if all consumers of the docker image prefer this 🤔 so no idea what to do with it. Can you maybe host this by yourself?

halkeye:
Yea I can't say either. I generally think rootless docker images are good practice. 

I believe I have a docker image created whenever a new release happens, so I'm not desperate, but being lazy it's nicer upstream :)

PR #925, opened 2024-03-04
Point to correct path for auth.html so completing sso login flow no longer 404s

halkeye:
The path is just /auth.html in the docker image, not sure if its a different path for other installation methods


krille-chan (approved):
thank you so much

PR #969, opened 2024-04-01
chore: do not use static openssl

Decodetalkers:
I close this option, to make it compilable on archlinux.

see the readme
https://github.com/simolus3/sqlite3.dart/blob/main/sqlcipher_flutter_libs/README.md

I do not think the option is needed  on linux machine

PR #986, opened 2024-04-04
feat: use dynamic linking for OpenSSL on Linux

TheOneWithTheBraid:
- dynamically link OpenSSL
- OpenSSL is _really_ preinstalled on any Linux machine
- shrinks the app size on Linux
- drops dependency on an OpenSSL static installation on development machines. This makes packaging for many distributions easier.

Since Version 0.6.1, `package:sqlcipher_flutter_libs` finally supports dynamic linking for OpenSSL. We can finally make use of it.

There is no breaking behavior about this.

Related patches in Arch Linux package upstream :
- https://aur.archlinux.org/cgit/aur.git/tree/openssl-dynamic-link.patch?h=fluffychat
- https://aur.archlinux.org/cgit/aur.git/tree/openssl-dynamic-link.patch?h=fluffychat-git

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -13,6 +13,9 @@ set(APPLICATION_ID "chat.fluffy.fluffychat")
 # versions of CMake.
 cmake_policy(SET CMP0063 NEW)
 
+# dynamically link OpenSSL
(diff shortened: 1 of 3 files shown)


TheOneWithTheBraid:
Ah, I just saw #969 ... Looks like someone already ported over that patch.

PR #1056, opened 2024-04-24
chatlist is laggy, speed a little up

dG94Cg:
Room.hasNewMessages invoke some heavy CPU function like FromJson (noticed from flutter run -d android --profile, CPU profiler), so using FutureBuilder to wrap it, and the room.isDirectChat will query all rooms which is slow too but it won't change at one render frame so make it as a temporary variable which may help.

> ChatListItem still will cost 80ms sometime, hope it will be solve latter.

### DIFF
--- a/lib/pages/chat_list/chat_list_item.dart
+++ b/lib/pages/chat_list/chat_list_item.dart
@@ -86,7 +86,7 @@ class ChatListItem extends StatelessWidget {
         borderRadius: BorderRadius.circular(AppConfig.borderRadius),
         clipBehavior: Clip.hardEdge,
         color: backgroundColor,
-        child: FutureBuilder(future: () async {return Future.value(room.hasNewMessages);} (),
+        child: FutureBuilder(future: Future.microtask(() => room.hasNewMessages),
         initialData: false, 
         builder: (context, hasMoreMessage)  {
           final hasNewMessages = hasMoreMessage.data ?? false;
--- a/lib/pages/chat_list/chat_list_item.dart
+++ b/lib/pages/chat_list/chat_list_item.dart
@@ -67,11 +67,6 @@ class ChatListItem extends StatelessWidget {
     final lastEvent = room.lastEvent;
     final ownMessage = lastEvent?.senderId == Matrix.of(context).client.userID;
     final unread = room.isUnread || room.membership == Membership.invite;
-    final unreadBubbleSize = unread || room.hasNewMessages
-        ? room.notificationCount > 0
-            ? 20.0
-            : 14.0
-        : 0.0;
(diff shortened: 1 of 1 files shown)


lib/pages/chat_list/chat_list_item.dart
```diff
@@ -90,7 +86,16 @@ class ChatListItem extends StatelessWidget {
         borderRadius: BorderRadius.circular(AppConfig.borderRadius),
         clipBehavior: Clip.hardEdge,
         color: backgroundColor,
-        child: FutureBuilder(
+        child: FutureBuilder(future: Future.microtask(() => room.hasNewMessages),
```
krille-chan, line 89: I don't get why this should make anything faster. If I understand this correctly this does not do this in a different thread, just in the event loop on the same thread. On the other hand, using a FutureBuilder here adds another non-const stateful widget to the tree which builds at least 2 times.

As far as I can see `room.hasNewMessages` might need a little bit of CPU time but isn't it enough to calc this ones like you did with `isDirectChat`?

dG94Cg: oh, it won't be faster in calculate. the reason of moving it into other Event Thread is it will block UI thread, the UI thread should be complete in 30ms, better in 16ms, ``room.hasNewMessages`` will take more than 30ms in most case.

as for ``isDirectChat``, it take a few CPU time too, but it is acceptable if it was only invoke for one time.

2024-04-27-144821_779x184_scrot

this is the screenshot of scroll chat-list-item in chat-list widget, it show the performance, if we move the ``room.hasNewMessages`` into FutureBuilder which invoke this API in eventloop, it won't block the UI thread which build widget. 
> PS:  there is a RED block but i think it is still acccept able which build take below 60ms which means we still got  31 FPS after this frame. and it is not usually happen.

2024-04-27-150108_669x301_scrot

2024-04-27-150222_709x712_scrot

this is the cpu time of invoke ``Room.hasNewMessages``, which it make drawFrame longer than it was suggested to, it take 72ms, which will block the render thread make user uncomfortable. if we move it into EventLoop, it would be smooth.

krille-chan: > oh, it won't be faster in calculate. the reason of moving it into other Event Thread is it will block UI thread

That is not the case when using Future.microtask. This still runs on the UI thread and does not unblock anything.

As I said before if you use the FutureBuilder widget, this will render the whole RoomListItem twice so this sounds like it gets even worse.

However the Room.isDirectChat change sounds like a good idea. I would like to merge this on it's own if you change your PR in this way

krille-chan:
I've now implemented some improvements on my own. @dG94Cg I'm still waiting for your response :-/

PR #1077, opened 2024-04-29
archive button

waqasahmedse76:


krille-chan:
@waqasahmedse76 445 commits? This was probably not intended I guess

MagicRB:
This PR makes references to something called Pangea Chat, could be this? https://app.pangea.chat/

MagicRB:
Theyve been active since 2019 according to the google play storw and i cannot find the current sourcecode anywhere. As fluffychat is licensed under AGPL only they have been breaking the license for 4 years.

PR #1091, opened 2024-05-07
Welcome chat abbr

Kelrap:
Changes name of automatically generated welcome chat to use an acronym of the class/exchange name, instead of the full name


Kelrap:
Wrong repository

PR #1130, opened 2024-05-23
deb: added debian package building

throwException:


krille-chan:
Would it be okay if you do this in your own repo and importing FluffyChat to this? I don't really want to take over maintainer work for 100 Linux packaging formats

throwException:
> Would it be okay if you do this in your own repo and importing FluffyChat to this? I don't really want to take over maintainer work for 100 Linux packaging formats

That works for me.

throwException:
@krille-chan Here is the package creation repo: https://github.com/throwException/fluffychat-debian

Unfortunatly the produced version doesn't work because of https://github.com/krille-chan/fluffychat/issues/1131

PR #1135, opened 2024-05-26
Various fixes for quick switching

UserJHansen:
There are a few cases where the global client is used in the chat window instead of the quick selected client, which this corrects.


krille-chan:
Thanks for the contribution. Please open only one Pull Request per change and make sure your commits use conventional commit message format

UserJHansen:
This is a different change and so I thought it would do better in a different pull request, so this one could be merged first. This one covers the chat elements, the other request covers the bundle management. 

To fix the commits, is it possible that you squash and merge instead of merging?

UserJHansen:
This should be good to merge again?

PR #1136, opened 2024-05-26
Fix the Bundle menu

UserJHansen:
Fixes the bundle menu so that it correctly updates when the user makes changes. 

As a small side effect these are all contained in PopupMenuItem so it looks slightly different:

image

vs

image


UserJHansen:
I'm going to open a new one with the git history fixed

PR #1137, opened 2024-05-26
Fix issue with bundle menu

UserJHansen:
Fixes the bundle menu so that it correctly updates when the user makes changes. 

As a small side effect these are all contained in PopupMenuItem so it looks slightly different:

image

vs

image

### DIFF
--- a/lib/pages/chat_list/client_chooser_button.dart
+++ b/lib/pages/chat_list/client_chooser_button.dart
@@ -19,14 +19,7 @@ class ClientChooserButton extends StatelessWidget {
 
   List<PopupMenuEntry<Object>> _bundleMenuItems(BuildContext context) {
     final matrix = Matrix.of(context);
-    final bundles = matrix.accountBundles.keys.toList()
-      ..sort(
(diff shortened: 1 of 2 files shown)


lib/pages/chat_list/client_chooser_button.dart
```diff
@@ -89,66 +82,93 @@ class ClientChooserButton extends StatelessWidget {
...
+                            icon: const Icon(Icons.edit_outlined),
+                            onPressed: () async {
+                              await controller.editBundlesForAccount(
+                                client.userID,
+                                bundle,
+                              );
+                              await client.oneShotSync();
+
+                              if (context.mounted) {
+                                setState(() {});
```
krille-chan: This does only update the GUI if the change was done on this device. What if the change was triggered on a different device? We should use a StreamBuilder here and listen to the account data changes instead

UserJHansen: I'll get that written now

krille-chan:
@UserJHansen can you please sign your commits?

UserJHansen:
> @UserJHansen can you please sign your commits?

All new commits are now signed

Issue #957, opened 2024-03-28
reply gets triggered when using gestures.


cyrneko:
### Bug Description

The swipe-to-reply function gets triggered when trying to use Android's back gesture.

### Steps to Reproduce

1. Enter any chat
2. Try to use the back gesture
3. Reply gets triggered

See video:

https://github.com/krille-chan/fluffychat/assets/58338843/dd95d11e-19df-4fc0-bdb0-d9fd0e18efc9

### Expected Behavior

The back gesture shouldn't trigger a reply.

### App Version

1.18.0, Google Play

### Additional Platform Information

Poco F3, Android 13; LineageOS for microG

### Additional Context

One proposal I can make is a swipe threshold, where the reply action is only triggered after the message bubble has moved X amount of pixels. Not sure if Fluffy does this already 🥴

Mek101:
I can reproduce!

SteffoSpieler:
I have the same issue. Also happens with the swipe to archive action in the chat list.

poettig:
Happens for me regularly as well, also left rooms because of it (and failure to press "cancel" instead of "leave"). With swipe-to-archive, it can even happen that both the back/exit gesture an the archive popup get triggered, causing a surprise popup the next time you open flufflychat :D

Issue #891, opened 2024-02-11
option to change swipe-to-reply behaviour


pbsds:
### Feature Description

Swiping right-to-left on chat bubbles to reply interferes with the swipe-from-edge-of-screen-to-go-back feature on Android. Adding the option to either disable this behaviour, changing the direction, or tune the sensitivity would be great

### Rationale

I keep replying to people when I try to exit a chat

Mm2PL:
I keep replying when I try to scroll which is really annoying. I wish I could turn this feature of completely.

@github-actions, you hunk of code, do not close this feature request its important to me!

Janavanwalde:
Fixed. Can now be turned off in the settings. :)

PR #1159, opened 2024-06-03
feat: add option to configure reply swipe direction

MrSpoony:
I'd really love there to be an option to reply by swiping from left to right. So I added a small pull request with the corresponding changes.

There is now an option to change the swipe direction in the settings.

:white_check_mark: Tested on Android

Issues that "kinda" reference this: #891 and #957

### DIFF
--- a/assets/l10n/intl_en.arb
+++ b/assets/l10n/intl_en.arb
@@ -180,6 +180,8 @@
   },
   "sendTypingNotifications": "Send typing notifications",
   "@sendTypingNotifications": {},
+  "swipeRightToLeftToReply": "Swipe right to left to reply",
+  "@swipeRightToLeftToReply": {},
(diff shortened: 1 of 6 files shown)


krille-chan (approved):
Thanks, that was very easy to review and looks clean. Normally I am against too many options, but as FluffyChat runs on so many platforms, it might make sense to make this configurable to solve the problem, when the swipe direction collides with a platform specific swipe gesture

Mek101:
> Thanks, that was very easy to review and looks clean. Normally I am against too many options, but as FluffyChat runs on so many platforms, it might make sense to make this configurable to solve the problem, when the swipe direction collides with a platform specific swipe gesture

At least on android the backwards swipe can be triggered on both sides, so I'm not sure how effective this workaround could be

PR #1239, opened 2024-07-22
Add RTL support (Bidi)

### LINKED ISSUE #63: Add bidi support
#### Description
Messengers are among those apps in which people with different languages (RTL/LTR) might make content. It is also possible people make mixed text on RTL and LTR. Therefore it is crucial to support bidirectional (bidi) text support in such apps. currently FluffyChat (at least in version 0.30.0 on Gnu/Linux Kubutnu 20.04) doesn't support it. As the result, text are shown in LTR direction regardless of in what language they are written.

#### To Reproduce
Simple! Just write some RTL text (from Persian, Arabic, Urdu,..) It would be better if it be mixed. I would sugget to write this text: "یک two سه four". If you see it in this order, it is wrong. It is now LTR since it is surrounded with LTR text. But if it be an independent body of text, it should be opposite. You may check the following screenshot for reference:

image

#### Additional information:
- Device: Laptop
- OS and OS version: Kubuntu 20.04
- Installed version of FluffyChat: 0.30.0 (from flathub)
- (Android only) Which store are you using: -
- (Android only) Are Google Services available: -

/issue fluffy

Amjad50:
Add Bidi support identification for text elements and dynamically change the direction of the text.

Fixes #63 

- [x] Android
- [ ] iOS
- [x] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat_app_bar_list_tile.dart
+++ b/lib/pages/chat/chat_app_bar_list_tile.dart
@@ -3,6 +3,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_linkify/flutter_linkify.dart';
 
 import 'package:fluffychat/config/app_config.dart';
+import 'package:fluffychat/utils/text_direction.dart';
 import 'package:fluffychat/utils/url_launcher.dart';
 
 class ChatAppBarListTile extends StatelessWidget {
@@ -33,7 +34,7 @@ class ChatAppBarListTile extends StatelessWidget {
             child: Padding(
               padding: const EdgeInsets.symmetric(horizontal: 4.0),
               child: Linkify(
-                text: title,
+                text: title.bidiFormatted,
                 options: const LinkifyOptions(humanize: false),
                 maxLines: 1,
                 overflow: TextOverflow.ellipsis,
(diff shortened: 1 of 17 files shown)


lib/pages/chat/chat_app_bar_list_tile.dart
```diff
@@ -33,7 +34,7 @@ class ChatAppBarListTile extends StatelessWidget {
             child: Padding(
               padding: const EdgeInsets.symmetric(horizontal: 4.0),
               child: Linkify(
-                text: title,
+                text: title.bidiFormatted,
```
krille-chan, line 37: Now every time I want to add a text somewhere, I need to think about adding the `.bidiFormatted` suffix? I would prefer a solution, which does not add more complexity

krille-chan (changes requested):
Thanks for the contribution, but this Pull Requests currently adds too much complexity, which I cannot maintain in the long term

Amjad50:
Thanks @krille-chan for the review. You are right, I think a better solution is needed here. let me see what can be done to improve this. 
The main issue is having direction control that is not tied to the system locale

waifucon:
Any progress? @Amjad50.

Amjad50:
> Any progress? @Amjad50.

Not really, I forgot about this as I have moved on, probably won't finish this.
if you are interested you can continue from where I left off

waifucon:
> Not really, I forgot about this as I have moved on, probably won't finish this. if you are interested you can continue from where I left off

Not a developer really. @krille-chan. Should look for some another maintainer for this?

PR #1247, opened 2024-07-26
Save custom emote reaction event keys as URI to render them

ggurdin:
Hello! It looks like the rendering logic for custom emote reactions is expecting the key in the reaction event to start with "mxc://", but currently custom emote reaction keys are being saved as their emote name instead. This change would parse those names into image URIs before sending so the keys would be saved in the "mxc://" format.

- [x] Android
- [ ] iOS
- [x] Browser (Firefox based)
- [x] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -455,18 +455,46 @@ class ChatController extends State<ChatPageWithRoom>
     var parseCommands = true;
 
     final commandMatch = RegExp(r'^\/(\w+)').firstMatch(sendController.text);
-    if (commandMatch != null &&
-        !sendingClient.commands.keys.contains(commandMatch[1]!.toLowerCase())) {
-      final l10n = L10n.of(context)!;
-      final dialogResult = await showOkCancelAlertDialog(
-        context: context,
-        title: l10n.commandInvalid,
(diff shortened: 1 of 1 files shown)


krille-chan (changes requested):
Thank you so much for your contribution. So if I understand this correctly, this converts `/react :custom-emote-name` into a `/react mxc://uri-of-custom-emote` while the mxc uri is coming from the room image pack. Seems to make sense.

But it looks like you are applying this by changing the text in the textfield before processing it. For this case it seems to me that it might be more clean to apply this behavior in the [Matrix Dart SDK](https://github.com/famedly/matrix-dart-sdk/) instead. Probably in this place: https://github.com/famedly/matrix-dart-sdk/blob/main/lib/src/utils/commands_extension.dart#L154

ggurdin:
Thanks for your reply @krille-chan! I have opened a PR over there (https://github.com/famedly/matrix-dart-sdk/pull/1894) and am closing this one.

denalena:
seems like it gets merged in matrix-dart-sdk soon :) https://github.com/famedly/matrix-dart-sdk/pull/1894

PR #1386, opened 2024-10-04
Minor UI fixes

k3-cat:

- [ ] ~~Every new feature or change of the design/GUI is linked to an approved design proposal in an issue~~
- [ ] ~~Every new feature in the app or the build system has a strategy how this will be tested and maintained from now on for every release, e.g. a volunteer who takes over maintainership~~

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

Sorry for the missing test on iOS, I couldn't build for it.

### DIFF
--- a/lib/config/themes.dart
+++ b/lib/config/themes.dart
@@ -1,3 +1,4 @@
+import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
@@ -72,6 +73,10 @@ abstract class FluffyThemes {
       useMaterial3: true,
       brightness: brightness,
       colorScheme: colorScheme,
+      cupertinoOverrideTheme: const CupertinoThemeData(
+        // This is required by adaptive_dialog
+        textTheme: CupertinoTextThemeData(),
+      ),
       textTheme: PlatformInfos.isDesktop
           ? brightness == Brightness.light
               ? Typography.material2018().black.merge(fallbackTextTheme)
--- a/lib/pages/chat/input_bar.dart
+++ b/lib/pages/chat/input_bar.dart
@@ -461,6 +461,10 @@ class InputBar extends StatelessWidget {
                 final data = content.data;
                 if (data == null) return;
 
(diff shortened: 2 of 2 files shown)


lib/pages/chat/input_bar.dart
```diff
@@ -461,6 +461,10 @@ class InputBar extends StatelessWidget {
                 final data = content.data;
                 if (data == null) return;
 
+                if (content.mimeType.startsWith("image/")) {
+                  onSubmitImage!(data);
```
krille-chan, line 465: I use this feature mostly with the GBoard when sending a sticker or GIF. Having a dialog every time would be just overhead IMO

lib/config/themes.dart
```diff
@@ -72,6 +73,10 @@ abstract class FluffyThemes {
       useMaterial3: true,
       brightness: brightness,
       colorScheme: colorScheme,
+      cupertinoOverrideTheme: const CupertinoThemeData(
+        // This is required by adaptive_dialog
```
krille-chan, line 77: Can you explain here, why it is required?

k3-cat: The issue:
image

The fix:
I thought this should be a common issue, so I've searched if there are any instructions.
and...
The change is guided by the readme of `adaptive_dialog` https://pub.dev/packages/adaptive_dialog#the-input-text-color-same-with-backgound-when-using-cupertinotextinputdialog

lib/pages/chat/input_bar.dart
k3-cat: Yeah... that will be very inconvenient....

But I think media is a bit expensive for the matrix. Currently both Android and web can upload images by pasting. So without visual confirmation, users might accidentally upload contents either they don't want or in its original quality (average screenshot takes about 7M, and in most case it's unnecessary to keep the original quality). Thus, putting more burden on home-level servers.

What if we only show the dialog for images larger than a certain size (e.g. 1M)?

PR #1447, opened 2024-11-01
feat: Better wallpapers with blur and opacity sliders and improved styles page

krille-chan:


Archenoth:
I know this is an old PR, but I just wanted to say, I really appreciate this being back, and also super-appreciate effort that went into implementing it!

Thank you!

Issue #1289, opened 2024-08-19
24-hour time format


eandersons:
### Feature Description

A setting to switch between time formats - AM/PM or 24-hour. Or, if possible, use time format based on system settings.

### Rationale

Currently time in chats is displayed in AM/PM time format, and it is not possible to switch to 24-hour format.

According on Wikipedia article [24-hour clock](https://en.wikipedia.org/wiki/24-hour_clock):

> This system, as opposed to the [12-hour clock](https://en.wikipedia.org/wiki/12-hour_clock), is the most commonly used time notation in the world today [..]

Mikaela:
Kind of related to https://github.com/krille-chan/fluffychat/issues/334

jilotta:
Second this. I wanted to recommend fluffychat to my not so tech-savvy friends but the am/pm time is weird and unreadable.

czAdamV:
This already seems to work on Linux, I'm using 24-hour time locale and FluffyChat displays the time correctly, but it doesn't work on Android.

eandersons:
> This already seems to work on Linux, I'm using 24-hour time locale and FluffyChat displays the time correctly, but it doesn't work on Android.

For me in FluffyChat 1.21.2 (installed from Flatpak) on ArchLinux with KDE Plasma time is displayed as AM/PM.

jilotta:
Works for me on FluffyChat 1.21.2 from AUR. Maybe it's because of flatpak sandboxing?

Porkepix:
Last release now have 24H time format for Android \o/
If Android was the last platform lacking support, this can be closed I guess (dunno about the status on other platforms).

eandersons:
Web (both fluffychat.im/web and self-hosted) and desktop (installed from Chaotic AUR) version (all 2.2.0)  still have 12 hour time format.

PR #1457, opened 2024-11-03
chore: use time format based on system settings

dawidd6:
Closes: #1289 

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


lib/utils/date_time_extension.dart
```diff
@@ -28,7 +28,8 @@ extension DateTimeExtension on DateTime {
 
   /// Returns a simple time String.
   String localizedTimeOfDay(BuildContext context) =>
-      L10n.of(context).alwaysUse24HourFormat == 'true'
+      MediaQuery.alwaysUse24HourFormatOf(context) ||
+              L10n.of(context).alwaysUse24HourFormat == 'true'
```
JAicewizard, line 32: Looking at the documentation (https://api.flutter.dev/flutter/widgets/MediaQueryData/alwaysUse24HourFormat.html) this seems to need different behavior depending on the platform. On android `MediaQuery.alwaysUse24HourFormatOf` should fully dictate the decision, but on IOS it should only force it on when `true` but use `L10n.of(context).alwaysUse24HourFormat` when `false`

kaanelloed: @JAicewizard But it's still handle both cases, no ?

In android, from my understanding `L10n.of(context).alwaysUse24HourFormat` will always be false, so only `MediaQuery.alwaysUse24HourFormatOf(context)` will apply.

In IOS, if one or both are true, it'll use the 24h format.

Porkepix:
Following merge of https://github.com/krille-chan/fluffychat/pull/2179, is this PR still needed?

dawidd6:
I guess not

PR #1488, opened 2024-11-17
fix: dont use thumbnails for emoticons

### LINKED ISSUE #1071: Animated emojis do not work
### Bug Description

FluffyChat does not support animated emojis because it currently loads them with the following URL format:

```https://matrix-client.matrix.org/_matrix/media/v3/thumbnail/matrix.org/<ID>?width=32&height=32&method=scale&animated=true```

However, the actual URL format should be:

```https://matrix-client.matrix.org/_matrix/media/r0/download/matrix.org/<ID>```

This loads the full resolution emoji and keeps the MIME type for animated emotes (APNG, WEBP, etc)

### Steps to Reproduce

1. Create an animated emoji using nheko or cinny
2. Send a message with the emoji
3. Look on FluffyChat and the emoji will not be animated

### Expected Behavior

The emoji should be animated.

### App Version

1.19.2

Linux, Android, iOS, Windows, and web

### Additional Context

CC @krille-chan

marekvospel:
In this pull request I went over places where `MxcImage` is used and in case of custom emojis set `isThumbnail` to false, as the thumbnail generation causes emoticons to lose transparency and in some cases overall animations. (more info in #1071)

The only place I'm worried about is `html_message.dart`, where I'm using the following snippet: `isThumbnail: (actualWidth * actualHeight) > (256 * 256)`. Maybe using something the `data-mx-emoticon` property would be more fitting?

Fixes #1071

- [X] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`
- [X] The commit message uses the format of [Conventional Commits](https://www.conventionalcommits.org)
- [X] The commit message describes what has been changed, why it has been changed and how it has been changed
- [X] ~~Every new feature or change of the design/GUI is linked to an approved design proposal in an issue~~
- [X] ~~Every new feature in the app or the build system has a strategy how this will be tested and maintained from now on for every release, e.g. a volunteer who takes over maintainership~~

- [ ] Android
- [ ] iOS
- [X] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -272,14 +272,18 @@ class ImageExtension extends HtmlExtension {
     final width = double.tryParse(context.attributes['width'] ?? '');
     final height = double.tryParse(context.attributes['height'] ?? '');
 
+    final actualWidth = width ?? height ?? defaultDimension;
+    final actualHeight = height ?? width ?? defaultDimension;
+
     return WidgetSpan(
       child: SizedBox(
-        width: width ?? height ?? defaultDimension,
-        height: height ?? width ?? defaultDimension,
+        width: actualWidth,
+        height: actualHeight,
         child: MxcImage(
           uri: mxcUrl,
-          width: width ?? height ?? defaultDimension,
-          height: height ?? width ?? defaultDimension,
+          width: actualWidth,
+          height: actualHeight,
+          isThumbnail: (actualWidth * actualHeight) > (256 * 256),
         ),
       ),
     );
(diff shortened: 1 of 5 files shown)


krille-chan:
@marekvospel looks good to me :) thank you very much for your contribution. It looks like your commit is not correctly signed yet. Can you sign your commit? According to the policy only signed commits can go into the main branch. More info on that you can find here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

marekvospel:
> @marekvospel looks good to me :) thank you very much for your contribution. It looks like your commit is not correctly signed yet. Can you sign your commit? According to the policy only signed commits can go into the main branch. More info on that you can find here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Will do, I just forgot the `-S` flag. 😅

PR #1541, opened 2025-01-03
chore: Update Mastodon Link in README.md

DivineDominion:
I also checked the famedly.com link and there's no product like server hosting at all for end-users (it's for medicinal staff), but I left it in :)


krille-chan (approved):
Thank you so much 👍😊

Issue #1602, opened 2025-02-10
Remove option to create nameless spaces


Kelrap:
When a new space is created, a name is not required. A nameless space is given a placeholder name of 'Empty chat'. 'Empty chat' is also used for nameless groups, which may lead to confusion over the difference between spaces and groups. 

1. Navigate to chat list view
2. Select '+ Chat' button at the bottom of the chat list
3. Select 'Create group'
4. Toggle 'Space' 
5. Select 'New space'
6. A new space is created, with the placeholder name 'Empty chat'

wcjord:
@ggurdin hit this one please!

PR #1641, opened 2025-02-21
fix: prevent users from creating group chats or spaces with empty names

ggurdin:

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

Hello! This PR is related to issue #1602 
I'm not sure if the option to create rooms without names is there on purpose (if it is, please let me know), but it can be a confusing user experience. This PR prevents users from creating rooms with empty names and gives them a warning message when they try to do so.

### DIFF
--- a/lib/pages/new_group/new_group.dart
+++ b/lib/pages/new_group/new_group.dart
@@ -2,6 +2,7 @@ import 'dart:typed_data';
 
 import 'package:flutter/material.dart';
 
+import 'package:flutter_gen/gen_l10n/l10n.dart';
 import 'package:go_router/go_router.dart';
 import 'package:matrix/matrix.dart' as sdk;
 import 'package:matrix/matrix.dart';
@@ -112,6 +113,12 @@ class NewGroupController extends State<NewGroup> {
     final client = Matrix.of(context).client;
 
     try {
+      if (nameController.text.trim().isEmpty &&
+          createGroupType == CreateGroupType.space) {
+        setState(() => error = L10n.of(context).pleaseFillOut);
+        return;
+      }
+
       setState(() {
         loading = true;
         error = null;


lib/pages/new_group/new_group.dart
```diff
@@ -112,6 +113,11 @@ class NewGroupController extends State<NewGroup> {
     final client = Matrix.of(context).client;
 
     try {
+      if (nameController.text.isEmpty) {
+        setState(() => error = L10n.of(context).pleaseEnterName);
```
krille-chan: There is already a translation which might fit here:

```suggestion
        setState(() => error = L10n.of(context).pleaseFillOut);
```

krille-chan (changes requested):
I'm not sure about this. Couldn't it be desired to create unnamed chats? Those chats are then named after the users you share them with like "Group with Alice and Bob". Of course you can create a chat and then invite nobody but then this should be obvious to the user that the chat is empty so the name "Empty chat" makes sense

lib/pages/new_group/new_group.dart
ggurdin: Thanks, I have updated this!

ggurdin:
@krille-chan Yes, I think that makes sense, especially for group chats. Do you think it would make sense to apply this only when creating spaces, or would the same logic apply to spaces as well (i.e. "Space with Alice and Bob")?

ggurdin:
> > @krille-chan Yes, I think that makes sense, especially for group chats. Do you think it would make sense to apply this only when creating spaces, or would the same logic apply to spaces as well (i.e. "Space with Alice and Bob")?
> 
> Making a name mandatory for new spaces sounds like a good idea I would say :-) Let's proceed with this

Thank you! I have updated the changes

lib/pages/new_group/new_group.dart
```diff
@@ -112,6 +113,12 @@ class NewGroupController extends State<NewGroup> {
     final client = Matrix.of(context).client;
 
     try {
+      if (nameController.text.isEmpty &&
```
krille-chan:
```suggestion
      if (nameController.text.trim().isEmpty &&
```

Small suggestion: We should trim the entered text so that we don't allow just whitespaces as a name

ggurdin: Makes sense! I've updated this

krille-chan (approved):

PR #1733, opened 2025-03-30
feat: Haptic feedback while swiping to reply

xegim:
Adds light haptic feedback when swiping to reply, similar to what Whatsapp, Signal and Telegram do. The feedback is given during the swipe, when the threshold distance for the action is exceeded. Additionally, the movement speed of the element is doubled to better follow finger movement. Currently, the message element lags behind while being swiped because the swipe_to_action package does not handle this properly.

This PR relies on minor changes to swipe_to_action package. Ideally, one should make a PR to the upstream but it does not seem to be open for public contributions. As a workaround, a fork of the package is used. The changes to the package can be seen [here](https://github.com/xegim/swipe_to_action/compare/master...onThresholdCallback).

- [x] Every new feature in the app or the build system has a strategy how this will be tested and maintained from now on for every release, e.g. a volunteer who takes over maintainership

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


pubspec.yaml
```diff
@@ -83,7 +83,10 @@ dependencies:
   slugify: ^2.0.0
   sqflite_common_ffi: ^2.3.3
   sqlcipher_flutter_libs: ^0.6.1
-  swipe_to_action: ^0.3.0
+  swipe_to_action:
```
krille-chan, line 86: Thank you so much, this looks great. Could you maybe make sure the pull request gets merged into the upstream package so that we don't introduce another git import?

xegim: That'd be the preferred approach surely. I'm not sure how to contact the upstream devs though, they didn't make it easy. The repository host does not accept registrations and I'm not sure if the email in the commits is even a valid one. We'll see.

krille-chan: Maybe an easy fix would be to trigger the haptic feedback in the app code, when the action is triggered. As it is only one line this should be fine :)

xegim: Triggering the feedback on the action would remove the need to change the package, but in my opinion it would also make the feedback somewhat pointless. After all, the entire idea of the feedback is to tell the user _when_ to release the swipe. Unfortunately as far as I can tell, adding a new callback to the package is the only approach🤔. If it makes things more manageable, I could propose that even you could fork swipe_to_action and point the pubspec to that until someday we get this upstream?

xegim:
@krille-chan The upstream repository for swipe_to_action is no longer up, and has not been for many months now. Do you think we could then host it elsewhere for fluffychat and include the patches?

PR #1784, opened 2025-04-22
feat: support threads

anbraten:
image

# TODO
- [ ] show thread messages in separate view
- [ ] show button to open thread
- [ ] hide thread messages from default chat
- [ ] allow to write message in thread


krille-chan:
I guess this is the start of an effort to implement threads? Please be aware that changes to the GUI should follow an approved mock up in an issue first. Otherwise it could likely happen, that it gets declined.

PR #1832, opened 2025-05-05
fix: maintain state of text input dialog to prevent input being erased on keyboard submit

ggurdin:
Hello! I ran into a problem with some text input dialogs on mobile. Sometimes, the mobile keyboard text input action defaults to a submit button (I think this happens when maxLines is set to 1, but it could be triggered in other ways). When this happens, clicking the submit button on the mobile keyboard does not submit the dialog, and instead erases the current text in the dialog.

This is because the TextEditingController in the text input dialog is remade each time that the dialog rebuilds.

I tested with wrapping the dialog in a StatefulBuilder, but had no luck with maintaining the text input, so instead I turned the dialog into a StatefulWidget. This also enables the TextEditingController to be disposed, preventing it from building up in memory.

I also added an onSubmitted callback to the text input dialog that calls the same function also the submit button. It seems intuitive that these two actions should do the same thing.

- [ x ] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`
- [ x ] The commit message uses the format of [Conventional Commits](https://www.conventionalcommits.org)
- [ x ] The commit message describes what has been changed, why it has been changed and how it has been changed

- [ x ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/widgets/adaptive_dialogs/dialog_text_field.dart
+++ b/lib/widgets/adaptive_dialogs/dialog_text_field.dart
@@ -17,6 +17,8 @@ class DialogTextField extends StatelessWidget {
   final TextInputType? keyboardType;
   final int? maxLength;
   final bool autocorrect = true;
+  final FocusNode? focusNode;
+  final void Function(String)? onSubmitted;
 
   const DialogTextField({
     super.key,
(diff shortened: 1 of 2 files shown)


krille-chan:
Thanks, but this is already fixed in the main branch by just moving one line of code more above

PR #1971, opened 2025-06-19
Weblate conflict

lucasmz-dev:
Basically just a merge from Weblate excluding the English modifications

It seems like editing English strings triggers issues...? If FluffyChat would like to not allow editing the source, AKA english strings, this is possible by disabling "Edit base file". It's complicated because you don't want Weblate to fight you except in translation files.


lucasmz-dev:
If it is desired to block editing English strings, or letting Weblate end up mutating them, then disable 'Edit base file' in the "Translations" component inside the Translation settings sub-menu for that component.

lucasmz-dev:
I noticed some empty @<stringnamehere>: {}, so just empty stuff that Weblate seems to add.
I also noticed that you never seemed to play with the add-ons, https://hosted.weblate.org/changes/browse/fluffychat/translations/?action=61&action=60&action=62&user=&exclude_user=&period=
That might be the issue, there are add-ons that solve common issues like this, like cleanup empty strings.

lucasmz-dev:
I'm not familiar with the format here ofc! I mess more with Android, but still, worth looking into
Also sorry if you did, nothing shows up in history though at least

lucasmz-dev:
@krille-chan can you get a quick look of this? I wanna take advantage of my want to go messing with pt-BR in Weblate

krille-chan:
Okay thank you. I'll try it out :)

PR #1976, opened 2025-06-20
fix(notification/linux): default action also need to be handled

Decodetalkers:
default should not be ignored

https://specifications.freedesktop.org/notification-spec/1.3/basic-design.html

but there still be a problem

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: 'package:go_router/src/router.dart': Failed assertion: line 521 pos 12: 'inherited != null': No GoRouter found in context
```

update:

fixed. After read the example and document from flutter, the context.go in that place was a misused. the context to use should be from the build function, for example https://github.com/flutter/packages/blob/main/packages/go_router/example/lib/main.dart#L57. Here we use the context from the go_route, then it never work.. I think there maybe other places also have such kind of mistake..

seems not relate to my changes... it does not work for long time ago

- [ ] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`
- [ ] The commit message describes what has been changed, why it has been changed and how it has been changed

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/widgets/local_notifications_extension.dart
+++ b/lib/widgets/local_notifications_extension.dart
@@ -1,10 +1,9 @@
 import 'dart:io';
 
-import 'package:collection/collection.dart';
-import 'package:fluffychat/widgets/fluffy_chat_app.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 
+import 'package:collection/collection.dart';
 import 'package:desktop_notifications/desktop_notifications.dart';
 import 'package:matrix/matrix.dart';
 import 'package:universal_html/html.dart' as html;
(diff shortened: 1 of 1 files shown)


Decodetalkers:
here even do context.go("/room/settings"), it also throws error.. I do not know much about flutter route..

Decodetalkers:
with

```dart
            FluffyChatApp.router.go('/rooms/${event.room.id}');

```

it works.. but from the example, the context should work..

krille-chan (approved):
Thank you so much <3

Issue #1026, opened 2024-04-15
Separate ABIs for F-Droid


shuvashish76:
Currently the universal apk size on F-Droid is bit heavy (~80MB), request to split it by their architecture.

Coool:
@krille-chan, I won't be rude. I haven't see such mess with release builds of any app. When are planned to be released seperate[^seperate-releases] architecture builds? I hate universal[^uni] builds. And what's up with different build[^diff] versions of same v2.4.0 release? Currently viable option to save storage space is to install mystical Google Play version which supports seperate architectures.

[^seperate-releases]: [Individual architecture builds and GitHub releases](https://github.com/zulip/zulip-flutter/issues/1917) - Identical Zulip app issue

[^uni]: [Reason I hate universal builds](https://github.com/guardianproject/orbot-android/issues/1459#issuecomment-3445925410), [Other](https://github.com/spacecowboy/Feeder/issues/493)

[^diff]: Google Play has [v2.4.0 build 6204](https://play.google.com/store/apps/details?id=chat.fluffy.fluffychat), but GitHub - [v2.4.0 build 3546](https://github.com/krille-chan/fluffychat/releases/tag/v2.4.0)

krille-chan:
@Coool it's in your hands. Can you send a patch? Or otherwise if you are interested in having this fixed by developers you can support the project here and send in the message the issue number to increase the chance that this gets picked up: https://ko-fi.com/krille

Coool:
@krille-chan, how is set up release schedule currently? Do you use current release GitHub workflow action? What you would like  to see (improve) on current release schedule?

krille-chan:
> [@krille-chan](https://github.com/krille-chan), how is set up release schedule currently? Do you use current release GitHub workflow action? What you would like to see (improve) on current release schedule?

I have no idea how f-droid is doing it. I'm not maintaining this. Afaik f-droid is fetching the repo on every release tag and then builds by itself?

For playstore releases it's done here: https://github.com/krille-chan/fluffychat/blob/main/.github/workflows/release.yaml

shuvashish76:
ping @linsui

linsui:
Why do you ping me? There is already a PR for this. There is nothing I can do.

Coool:
[...]

PR #1978, opened 2025-06-21
build: split APKs by ABI to optimize size

Integral-Tech:
This commit splits APKs by ABI (arm64-v8a, armeabi-v7a & x86_64) to optimize APK size.

Closes: #1026

- [x] Every new feature in the app or the build system has a strategy how this will be tested and maintained from now on for every release, e.g. a volunteer who takes over maintainership

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/android/app/build.gradle.kts
+++ b/android/app/build.gradle.kts
@@ -1,5 +1,6 @@
 import java.util.Properties
 import java.io.FileInputStream
+import com.android.build.api.variant.FilterConfiguration.FilterType.*
 
 plugins {
     id("com.android.application")
@@ -86,4 +87,23 @@ android {
 
 flutter {
     source = "../.."
+}
+
+// From https://developer.android.com/build/configure-apk-splits#configure-APK-versions
+//
+// The following code creates a mapping that assigns a unique numeric value for each ABI
+// and overrides the output version code.
+val abiCodes = mapOf("x86_64" to 1, "armeabi-v7a" to 2, "arm64-v8a" to 3)
+
+androidComponents {
+    onVariants { variant ->
+        variant.outputs.forEach { output ->
+            val name = output.filters.find { it.filterType == ABI }?.identifier
+
+            val baseAbiCode = abiCodes[name]
+            if (baseAbiCode != null) {
+                output.versionCode.set(baseAbiCode + (output.versionCode.get() ?: 0) * 10)
+            }
+        }
+    }
 }


linsui:
For F-Droid users the client select the apk for them. For GitHub users you can publish the universal apk only.

Integral-Tech:
> For F-Droid users the client select the apk for them. For GitHub users you can publish the universal apk only.

@krille-chan Would you like to do that?

linsui:
@krille-chan We can also only do this on F-Droid side if it works for you. The version code will be different from your apks but that won't cause any problem.

Integral-Tech:
> @krille-chan We can also only do this on F-Droid side if it works for you. The version code will be different from your apks but that won't cause any problem.

@krille-chan @linsui Resolved :)

Integral-Tech:
@krille-chan Flutter 3.35.3 has been released which fixes the version code issue (https://github.com/flutter/flutter/issues/173917)
Could you please update Flutter to 3.35.3 and approve this pull request?

android/app/build.gradle.kts
```diff
@@ -87,3 +88,18 @@ android {
 flutter {
     source = "../.."
 }
+
+val abiCodes = mapOf("x86_64" to 1, "armeabi-v7a" to 2, "arm64-v8a" to 3)
+
+androidComponents {
```
krille-chan, line 98: Can you add comments where you have this code block from and explain what it does?

krille-chan (changes requested):
In some CI jobs we upload the apk file. I guess now we need to upload multiple files? Are we sure that this won't break because the apk file name might change?

android/app/build.gradle.kts
Integral-Tech: @krille-chan Resolved :)

Integral-Tech:
@krille-chan Building a universal APK is still supported :)
Would you like to publish the universal APK only in GitHub Release, or publish both split APKs and the universal APK?

linsui:
The code needs to be updated. See https://github.com/saber-notes/saber/blob/main/android/app/build.gradle.kts#L76-L85 for example.

android/app/build.gradle.kts
krille-chan: Why is this not the Flutter default? I don't want to differ from the default if there is no a very important reason to do it. Please add a motivation text to differ from the flutter default with an upstream issue in flutter to the comment here

Integral-Tech: > Why is this not the Flutter default? I don't want to differ from the default if there is no a very important reason to do it. Please add a motivation text to differ from the flutter default with an upstream issue in flutter to the comment here

According to [Android Developers Documentation](https://developer.android.com/build/configure-apk-splits#configure-APK-versions):

> By default, when Gradle generates multiple APKs, each APK has the same version information, as specified in the module-level build.gradle or build.gradle.kts file.

krille-chan: This doesn't answer my question

Integral-Tech: > This doesn't answer my question

I mean the default version code for multiple APKs is unrelated to Flutter. "Each APK has the same version information" applies to both native Android and Flutter apps.

PR #2028, opened 2025-07-22
Selectable code and pasting of images

MrFastDie:
I made a POC for being able to copy code blocks and pasting images. 

I had to also include https://github.com/1runeberg/code_highlight_view as a replacement for highlight, as the other one did not receive any updates in centuries. 

However, this is only tested on Linux as I do not have other platforms to test, and I don't really have a high interest on setting up cross-platform deployment.

As I really like Fluffychat as an alternative to Element that breaks every second breath you take, I'd like to add my comfort features. However, I do not really feel able to take Ownership of testing and maintaining these features as I basically don't have any experience within flutter.

- [ ] Every new feature or change of the design/GUI is linked to an approved design proposal in an issue **I changed the default theme for code as it's somehow a little broken / ugly with the replacement lib - I also included a hover button to copy the code**

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/input_bar.dart
+++ b/lib/pages/chat/input_bar.dart
@@ -1,15 +1,19 @@
+import 'dart:io';
+
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 
 import 'package:emojis/emoji.dart';
 import 'package:flutter_typeahead/flutter_typeahead.dart';
 import 'package:matrix/matrix.dart';
+import 'package:pasteboard/pasteboard.dart';
 import 'package:slugify/slugify.dart';
 
 import 'package:fluffychat/config/setting_keys.dart';
 import 'package:fluffychat/l10n/l10n.dart';
 import 'package:fluffychat/utils/markdown_context_builder.dart';
 import 'package:fluffychat/widgets/mxc_image.dart';
+import '../../utils/paste_intends.dart';
 import '../../widgets/avatar.dart';
 import '../../widgets/matrix.dart';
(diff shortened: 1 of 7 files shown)


lib/pages/chat/input_bar.dart
```diff
@@ -404,50 +408,58 @@ class InputBar extends StatelessWidget {
...
+      builder: (context, controller, focusNode) {
+        // Action when an image is detected
+        void handleImage(Uint8List bytes) {
+          if (onSubmitImage != null) {
+            onSubmitImage!(bytes);
+          } else {
             room.sendFileEvent(
-              file,
+              MatrixFile(
+                  bytes: bytes, mimeType: 'image/png', name: 'pasted.png'),
```
WebFreak001, line 419: The mimetype is wrong, on android it's passed from the OS as-is (maybe it could be jpg or others?), on windows it's definitely always BMP though. The underlying library is not very transparent about this and this just seems to be an implementation detail.

Also previously the dependencies for pasteboard were more incorporated in the build process, I don't know if this is relevant here to this PR though.

This seems to have been previously implemented but regressed here: https://github.com/krille-chan/fluffychat/commit/dda45f783f2bf699e2a6612cec611c4868c0a2eb
And the dependency you re-added was removed here: https://github.com/krille-chan/fluffychat/commit/933119a3085edecfe6131c182bc44ddd107061d1

PR #2083, opened 2025-08-01
fix: remove highlight when message is unpinned

### LINKED ISSUE #2078: Pin highlight stays after message is unpinned
### Bug Description

When you pin a message, it doesn't get any highlight or special marker in the chat by default, however if you click on the pinned message title displayed below the appbar to jump to it, it becomes highlighted in your primary color. Then, even after unpinning, it stays highlighted this color until you do the same to a different message.

https://github.com/user-attachments/assets/a5c7127f-0c15-44da-a4df-68cbf02e7c62

### Steps to Reproduce

1. Pin a message
2. Click on the pin-bar at the top to navigate to it
3. Unpin it
4. It'll still be highlighted until you do the same for another message, refresh, or leave and come back

### Expected Behavior

Once a message is pinned, it makes sense to change the highlight color overall and not just when it's navigated to. Also, once a message is unpinned, the color should go away. 

### App Version

2.0.0

Windows chrome

### Additional Context

_No response_

avashilling:
Removes highlight that you get from scrolling to a pinned message, when that message is unpinned. It also calls the unpinEvent function from clicking on the upper right pin icon, as well as the one on the pin bar so that the actions are consistent. Closes #2078 

- [ ] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1203,6 +1203,7 @@ class ChatController extends State<ChatPageWithRoom>
         pinnedEventIds.contains(selectedEventIds.single);
     if (unpin) {
       unpinEvent(selectedEventIds.single);
+      return;
     } else {
       pinnedEventIds.addAll(selectedEventIds);
     }
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -1184,6 +1184,11 @@ class ChatController extends State<ChatPageWithRoom>
     if (response == OkCancelResult.ok) {
       final events = room.pinnedEventIds
         ..removeWhere((oldEvent) => oldEvent == eventId);
+      if (scrollToEventIdMarker == eventId) {
+        setState(() {
+          scrollToEventIdMarker = null;
+        });
+      }
       showFutureLoadingDialog(
         context: context,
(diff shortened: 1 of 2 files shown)


lib/pages/chat/chat.dart
```diff
@@ -1197,7 +1202,7 @@ class ChatController extends State<ChatPageWithRoom>
     final unpin = selectedEventIds.length == 1 &&
         pinnedEventIds.contains(selectedEventIds.single);
     if (unpin) {
-      pinnedEventIds.removeWhere(selectedEventIds.contains);
+      unpinEvent(selectedEventIds.single);
```
krille-chan, line 1205: I don't understand this change. Why do you now call the unpinEvent method here?

edit: Ah to also trigger the remove highlight. But then you need to return here as well, because otherwise the code below the if else statement is still executed.

avashilling: I see, fixed!

Issue #293, opened 2023-07-15
Scrolling is very stuttery/laggy


krille-chan:
#### Description

Scrolling both encrypted and unencrypted messages, room lists and settings has a lot of stutter, which I wouldn't expect from a device as new as a Pixel 6.

Other people seem to be having the same issue in the GrapheneOS offtopic chat (https://matrix.to/#/!tgbPIxPlvGiwpsGJuu:matrix.org/$EOE_mhTtzAcs6uOogf9KgUGyxUiLF-4Xl3OeXHTFoxE?via=grapheneos.org&via=matrix.org&via=tchncs.de).

I couldn't find an exact issue, but it could be related to #511 except not quite as extreme.
Another similar issue is #650 but as suggested that may just be an issue with the Linux version.

#### To Reproduce

Open app

#### Additional information:
- Device: Google Pixel 6
- OS and OS version: GrapheneOS 12.0
- Installed version of FluffyChat: 1.2.0 (issue present for multiple versions before)
- (Android only) Which store are you using: F-Droid, although I tried the play version and the issue persists
- (Android only) Are Google Services available: No

imnotlxy:
On my Pixel 9 Pro, average FPS is 115. Most frames are rendered within 8ms but there are some extremely slow frames. UI rebuild contributes the most to these laggy frames. I will look into this.

imnotlxy:
@krille-chan Could you reopen the issue?

saihaze:
Building `ChatListItem` and `ChatListHeader`  are slow. To make the chat list scroll smooth, it is necessary to figure out how to reduce build cost of these two widgets. Both of them generate deep build trees.

I will start with `ChatListItem` which has a terribly long `build` method. It should anyway be refactored.

PR #2105, opened 2025-08-09
fix: build `HtmlMessage` lazily

saihaze:
Related to https://github.com/krille-chan/fluffychat/issues/293, reducing jank on the chat page. This patch avoids parsing HTML in the `build()` method. Note that there is a UX change. Bubbles now resize themselves more frequently.

### Problems to be resolved:

 - [ ] When scrolling down, newly appeared bubbles resize themselves and thus push the older ones up. This is annoying.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -506,17 +506,25 @@ class HtmlMessage extends StatelessWidget {
     }
   }
 
+  // `dom.Element.html('')` won't work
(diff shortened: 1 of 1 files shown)


imnotlxy:
@krille-chan What do you think about the remaining problem?

PR #2115, opened 2025-08-12
Reduce UI janks of the chat list

saihaze:
https://github.com/krille-chan/fluffychat/issues/293

`calcLocalizedBodyFallback()` sometimes can be really slow and shouldn't be directly called in a build method. This is the main cause of UI being laggy on the chat list page.

### Criteria:

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat_list/chat_list_item.dart
+++ b/lib/pages/chat_list/chat_list_item.dart
@@ -277,18 +277,21 @@ class ChatListItem extends StatelessWidget {
                                             directChatMatrixId !=
                                                 room.lastEvent?.senderId),
                                       )
-                                    : null,
-                                initialData:
-                                    lastEvent?.calcLocalizedBodyFallback(
-                                  MatrixLocals(L10n.of(context)),
-                                  hideReply: true,
-                                  hideEdit: true,
-                                  plaintextBody: true,
-                                  removeMarkdown: true,
-                                  withSenderNamePrefix: (!isDirectChat ||
-                                      directChatMatrixId !=
-                                          room.lastEvent?.senderId),
-                                ),
+                                    : Future(
+                                        () => lastEvent
+                                            ?.calcLocalizedBodyFallback(
+                                          MatrixLocals(L10n.of(context)),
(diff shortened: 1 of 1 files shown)


lib/pages/chat_list/chat_list_item.dart
```diff
@@ -277,18 +277,21 @@ class ChatListItem extends StatelessWidget {
...
-                                  hideEdit: true,
-                                  plaintextBody: true,
-                                  removeMarkdown: true,
-                                  withSenderNamePrefix: (!isDirectChat ||
-                                      directChatMatrixId !=
-                                          room.lastEvent?.senderId),
-                                ),
+                                    : Future(
+                                        () => lastEvent
+                                            ?.calcLocalizedBodyFallback(
```
krille-chan, line 282: What is the point in using a FutureBuilder, when the method does not return a future? Also this still calls it in the build method

imnotlxy: It's like a workaround. `calcLocalizedBodyFallback()` is slow when the message body is complicated, so I just intended not to block the `build()` method.

JAicewizard: This makes sense, expensive computations should be kept out `build` as it causes jank. This pushes the computation onto the eventqueue allowing the build to finish and the frame to render

imnotlxy: > This makes sense, expensive computations should be kept out `build` as it causes jank. This pushes the computation onto the eventqueue allowing the build to finish and the frame to render

I think it's just a work-around. `calcLocalizedBodyFallback()` was designed to be fast and to server as a placeholder, but it is actually slow. I think it necessary to re-consider this API.

imnotlxy:
The patch is not good, closing it for a better solution.

PR #2212, opened 2025-09-17
fix: make time format user-configurable (12h/24h support)

tobi-jolly:
### What has been changed
- Updated `date_time_extension.dart` to allow users to choose between 12h and 24h time formats instead of forcing one.

### Why it has been changed
- Some users prefer 12h format while others prefer 24h. This change respects the user’s system or app setting.

### How it has been changed
- Used `use24HourFormat` check to decide format dynamically.
- Verified the fix on:
  - Android (emulator)
  - Browser (Chromium based, Microsoft Edge)
  - Browser (Chromium based, Chrome)

### Notes
- Code formatted with `dart format` and imports sorted with `dart run import_sorter:main --no-comments`.
-  Commit follows the [Conventional Commits](https://www.conventionalcommits.org) format.

### DIFF
--- a/lib/utils/date_time_extension.dart
+++ b/lib/utils/date_time_extension.dart
@@ -7,24 +7,16 @@ import 'package:fluffychat/utils/platform_infos.dart';
 
 /// Provides extra functionality for formatting the time.
 extension DateTimeExtension on DateTime {
-  bool operator <(DateTime other) {
-    return millisecondsSinceEpoch < other.millisecondsSinceEpoch;
-  }
+  bool operator <(DateTime other) =>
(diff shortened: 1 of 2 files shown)


krille-chan (changes requested):
This conflicts with the recent changes on https://github.com/krille-chan/fluffychat/commit/e584c22c8bf1c12d0170e3e4e53282a00bc355f4 which fixes the behavior on iOS. This now changes the behavior on iOS to always ignore the user pref. The commit message says this makes it user-configurable, while the user cannot configure anything here (I would vote against such a config anyway)

krille-chan:
Closed because of inactivity

PR #2235, opened 2025-09-30
Add direct chat functionality using URL parameters

aturkenov:
Maybe you already familiar with a wa.me/+7 (whatsapp) or t.me/+7 (telegram) links.
So this merge request starts direct chat with a recipient using http URL parameter `?direct=<matrix-user-id>`.

- [ ] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`
- [ ] The commit message describes what has been changed, why it has been changed and how it has been changed

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat_list/chat_list.dart
+++ b/lib/pages/chat_list/chat_list.dart
@@ -11,6 +11,7 @@ import 'package:go_router/go_router.dart';
 import 'package:matrix/matrix.dart' as sdk;
 import 'package:matrix/matrix.dart';
 import 'package:receive_sharing_intent/receive_sharing_intent.dart';
+import 'package:universal_html/html.dart' as html;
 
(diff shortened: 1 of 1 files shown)


krille-chan (changes requested):
I see some problems with this solution:

1. When refreshing the chat List for some reason, it would repeat the action, right?
2. We should verify that the user ID is a valid matrix ID
3. The query parameter ?direct is not very intuitive. Element uses a route for it.
4. Instead of directly start a chat we should display the profile dialog so that the user can verify the correct user there and then tap on "Start chat".
5. What happens if you are not logged in?

PR #2261, opened 2025-10-19
ci: Build only for arm64

krille-chan:


TotallyAvailable:
@krille-chan  ***Please*** remember to set the GITHUB_TOKEN permissions (& 'Secret Source') for **all** workflows to the ***absolute minimum*** necessary to avoid further use of **default write-all** tokens.

This message has **nothing** to do with this PR and should only be seen as a heads-up to reduce the risk of leaking highly privileged tokens in the future. Should this scenario not be of concern, ignore this message. Apologies again for using this way of reaching out.

PR #2406, opened 2025-12-05
feat: Set default window size on Linux to 1110x950

0rzech:

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

This is to give the app some breathing air on desktop Linux. Currently FluffyChat starts very small by default and at least on Gnome requires resizing each time to get a better usability.

### DIFF
--- a/linux/my_application.cc
+++ b/linux/my_application.cc
@@ -60,7 +60,7 @@ static void my_application_activate(GApplication* application) {
     gtk_window_set_title(window, "FluffyChat");
   }
 
-  gtk_window_set_default_size(window, 864, 720);
+  gtk_window_set_default_size(window, 1110, 950);
 
   g_autoptr(FlDartProject) project = fl_dart_project_new();
   fl_dart_project_set_dart_entrypoint_arguments(project, self->dart_entrypoint_arguments);


krille-chan (changes requested):
I'm not sure. Personally I prefer the smaller size. As this is user specific we should more aim to remember the window size before closing the app while the best approach might be to build this into the Flutter framework

0rzech:
This is how much information fits on FluffyChat mobile: 
FluffyChat mobile

And this is how much information fits on FluffyChat desktop Linux:
FluffyChat desktop Linux

IMHO, the difference is pretty significant and the desktop Linux variant is much less usable by default with how narrow the chat widget is and how little vertical space it has.

Also, the Windows variant defaults to 1280x720, judging by [main.cpp](https://github.com/krille-chan/fluffychat/blob/e80e3273eabc31101a9f26fc0df33f196433f372/windows/runner/main.cpp#L29C3-L29C24) at least, so it's even wider than what I propose here. To be fair, though, MacOS version seems to be 800x600 ([MainMenu.xib](https://github.com/krille-chan/fluffychat/blob/e80e3273eabc31101a9f26fc0df33f196433f372/macos/Runner/Base.lproj/MainMenu.xib#L335)) only.

Perhaps it would make sense to do a poll on the FluffyChat Matrix channel asking others about the default size preference (assuming you're open to changing it 😉)?

I agree that window resizing support in Flutter framework would be the best solution, but I wouldn't bet on them adding it any time soon.

I also found [window_manager](https://pub.dev/packages/window_manager) plugin which allows changing desktop window size. There are others too, but this one _seems_ to be the most actively managed. Using it, the app could dump its current window size and position to some config file and restore that on start. There may be a small risk of creating some interesting edge cases on different desktop environments and multi-monitor setups etc.

I fully respect it's your project and you can close this PR anyway. 😉

Issue #2176, opened 2025-08-31
List of user experience issues


DearFox:
A list of user experience issues from the perspective of a typical user.
Some of these problems have already been described in other issues, please note them here.

- Swiping through pictures like in TikTok, which is why on Android it is impossible to close a picture with the usual gesture or zoom normally

https://github.com/user-attachments/assets/7b25f628-ccdc-4d0d-9504-cb4297b90749

- When searching for new rooms, you can select only 1 server from which rooms will be searched

<img width="958" height="768" alt="Image" src="https://github.com/user-attachments/assets/cb3b8347-d16a-47ad-9566-354302b45435" />

It's worth adding a list of different servers so that you can search for rooms in different places.

<img width="364" height="743" alt="Image" src="https://github.com/user-attachments/assets/1fda3c22-0145-4a22-9b99-66910af592fa" />

- In the PC version, if there are several rooms when searching, you will not be able to scroll them in any way, scrolling only works with touch input

<img width="399" height="182" alt="Image" src="https://github.com/user-attachments/assets/116c920a-a986-45e1-bde9-3ab210ca0b9e" />

If you enable touch input simulation in your browser, everything will work.

https://github.com/user-attachments/assets/95249904-9d1e-41d9-bbd2-59119cdbd164

- Ctrl+v doesn't work for images, at least in the browser version. (and for other files I suppose)

- The browser version works in some atypical way, because of which you can't even access the interface elements, and it also causes problems with some extensions

I understand that this is a bit of a technical aspect, but for example the translation extension works strangely and not always correctly. Also, most likely, various kinds of tools for people with disabilities may not work quite correctly.

- On Android (it turns out not only that, in the browser version it also doesn't work, at least in firefox), when writing text, there is no spell check in the input line, words are not highlighted in red if there is an error in them

<img width="305" height="84" alt="Image" src="https://github.com/user-attachments/assets/bb28d192-28d6-4cc3-9121-5ecce8c12bdf" />

and how it should work:

<img width="221" height="70" alt="Image" src="https://github.com/user-attachments/assets/14ff812e-6a14-40e9-b219-40c6b9831d39" />

- Text selection on android can be a bit chaotic

<img width="574" height="1280" alt="Image" src="https://github.com/user-attachments/assets/9a58cffc-f422-48ed-b211-229dfff03a91" />

[...]

Issue #1632, opened 2025-02-18
Can't pinch to zoom reliably on images


CeruleanDerpo:
### Bug Description

Pinching to zoom on images rarely works as scrolling the images takes priority.
Zooming vertically has higher chances to work

### Steps to Reproduce

1. Open an image in a room
2. Pinch on it to zoom
3. Try both horizontally and vertically
4. Observe how it works slightly more reliably when pinching vertically

### Expected Behavior

The image zooms reliably regardless of pinch rotation

### App Version

1.24

### Additional Platform Information

Google Pixel 7a, Android 15 QPR1

### Additional Context

Here's a recording of the issue:

https://github.com/user-attachments/assets/290a142d-5df2-40cc-adec-fec800699ff3

sadorowo:
i also have this issue, a temporary workaround is to use vertical zoom - it works most of the times

l3u:
I'm facing this exact problem using FluffyChat 2.1.1 on my Motorola g42 (Android 13).

As long as there's only one image in the respective room, zooming works as expected. As soon as there are more, one can browse through the images by swiping up and down. And most of the time when I try to zoom, a swipe happens instead.

@sadorowo's workaround works (rotated by 90°), if I don't do the two-finger-stretch guesture vertically but horizontally, zooming works.

This is definitely a bug that should be fixed. I actually was just about to write a feature request "Please allow zooming into images" because I was sure this wasn't possible … thanks in advance for fixing this!

l3u:
I really have no idea about Android development, but couldn't one simply distinguish the one intend from the other by counting the fingers? One finger → scroll, two fingers → zoom?! 🤔 Just thinking …

Gibbz:
I'm having the issue too on android 16.
Can't get zoom working at all. I'll try different angles to see if any help.

Edit: seems to be if you get the time just right on both fingers you can pinch zoom. If your out by the tiniest amount it default to swiping between images. Hopefully a easing of the 2 finger gesture timing could be all that's needed to fix this?

CeruleanDerpo:
The issue is still present on the latest version

aquada:
I'm am also getting this, very sticky behaviour trying to zoom. Sometimes 2 hands, a finger of each, needed to push apart horizontally is the only way.

l3u:
[...]

PR #2434, opened 2025-12-17
fix: improve image viewer gestures (driver approach)

Gibbz:
Replaces gesture arena logic with a driver approach where InteractiveViewer handles all touches. Adds manual scroll driving, double-tap zoom, and reliable pinch-to-zoom transition.

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/image_viewer/image_viewer_view.dart
+++ b/lib/pages/image_viewer/image_viewer_view.dart
@@ -8,6 +8,7 @@ import 'package:fluffychat/utils/platform_infos.dart';
 import 'package:fluffychat/widgets/hover_builder.dart';
 import 'package:fluffychat/widgets/mxc_image.dart';
 import 'image_viewer.dart';
+import 'zoomable_image.dart';
 
 class ImageViewerView extends StatelessWidget {
   final ImageViewerController controller;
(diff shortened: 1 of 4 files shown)


Gibbz:
this is related to the issues:
https://github.com/krille-chan/fluffychat/issues/1632
https://github.com/krille-chan/fluffychat/issues/2176

lib/pages/image_viewer/zoomable_image.dart
```diff
@@ -0,0 +1,108 @@
+import 'package:flutter/material.dart';
+
+class ZoomableImage extends StatefulWidget {
```
krille-chan, line 3: Where do you have this code from? Is it from a package or an example from stackoverflow?

lib/pages/image_viewer/image_viewer_view.dart
```diff
@@ -74,8 +75,13 @@ class ImageViewerView extends StatelessWidget {
               KeyboardListener(
                 focusNode: controller.focusNode,
                 onKeyEvent: controller.onKeyEvent,
+                // We disable the intrinsic scrolling of the PageView because we want
+                // the [ZoomableImage] to drive the scrolling manually.
+                // This allows us to seamlessly switch between scrolling (1 finger)
+                // and zooming (2 fingers) without the Gesture Arena locking us out.
```
krille-chan, line 81: This all looks like a lot of workarounds for something that should be built into the framework. Are you sure this is the correct approach, that we maintain 174 new lines of code by ourself?

l3u:
Thank you so much for working on this. Zooming not working as expected is a really annoying bug …

But when we're at it, I would like to add another aspect: In virtually all apps, browsing through images is implemented as swiping left and right. Here, we swipe up and down to do that. Whilst one could argue that this is what we do to navigate through the chat history, it's not what one would expect. Maybe it would be worth considering changing this, too?

Thanks again!

lib/pages/image_viewer/image_viewer_view.dart
Gibbz: I had a dig around and found the main issue is that the viewer is in a scroll widget. As soon as the scroll moves a tiny amount, it takes control of the gestures.
I tried working around the issue, but couldn't get it to work. So this was a better solution. It's still not great and the image viewer experience is still frustrating.

Maybe it would be better to use a specific library designed for image gallery. This would reduce the burden of image viewer maintenance and usability. The following looks to be popular and easy to use.

https://github.com/bluefireteam/photo_view/

Issue #2221, opened 2025-09-20
[Android 15] Share feature have bottom button mostly covered by the navigation buttons


Porkepix:
### Bug Description

This is probably a regression with Android 15, I don't remember such issue on Android 14.
When using the share feature and having navigation buttons, those mostly cover the "Share" button, making it hard to use, see joined screenshot:

<details>
<summary>
Screenshot
</summary>

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/78fd9c5b-d66e-4dba-be80-52a86f396050" />

</details>

### Steps to Reproduce

Be on Android 15 and use navigation buttons
Try to use the share feature.

### Expected Behavior

Navigation buttons shouldn't cover the bottom button

### App Version

2.1.1

### Additional Platform Information

Android 15, Moto G84

### Additional Context

That's one place where I've seen the issue, maybe it exists elsewhere.

Gibbz:
Yes same issue

PR #2437, opened 2025-12-18
fix: android share screen 

Gibbz:
share screen previously was overlaping the navigation and notification bars

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/utils/show_scaffold_dialog.dart
+++ b/lib/utils/show_scaffold_dialog.dart
@@ -12,7 +12,7 @@ Future<T?> showScaffoldDialog<T>({
   required Widget Function(BuildContext context) builder,
 }) => showDialog<T>(
   context: context,
-  useSafeArea: false,
+  useSafeArea: true,
   builder: FluffyThemes.isColumnMode(context)
       ? (context) => Center(
           child: Container(


Gibbz:
Fix for this issue
https://github.com/krille-chan/fluffychat/issues/2221

lib/utils/show_scaffold_dialog.dart
```diff
@@ -12,7 +12,7 @@ Future<T?> showScaffoldDialog<T>({
   required Widget Function(BuildContext context) builder,
 }) => showDialog<T>(
   context: context,
-  useSafeArea: false,
```
krille-chan, line 15: This does change it for **all** showScaffoldDialog uses. Have you checked all of them? Maybe better make this configurable, keep the default to false and only for the share screen make it true?

Gibbz: Ah! I'll have a dig and see what other dialogs are using it then test them.

Gibbz: I have checked the uses of this dialog. they are all working fine. There are several forwarding and sharing screens using it.
(`image_viewer.dart`, `chat_list.dart`, and `chat.dart`)

Gibbz: ops not sure if i should use resolved or not

PR #2541, opened 2026-02-07
feat: implement UnifiedPush on Linux

### LINKED ISSUE #486: Feature request: Run in background on Linux
### Feature Description

Allow to run Fluffychat in background on Linux. This way it can also be used on Linux Phones.

### Rationale

Better intergration with Linux phones.

### Mockup

_No response_

### Additional Context

_No response_

TheSunCat:

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

I decided to give it a go myself, just following the UnifiedPush Flutter connector's implementation (got a lot of help from their devs). The app will try to register on launch (like on Android) and give a dismissable message if not. If there are no available distributors, it will fallback to the old local notifications, otherwise everything goes through UnifiedPush. I mirrored most of the implementation from `local_notifications_extension` to `push_helper`, though it is quite annoying we have a duplicated impl for this (using different libraries).

Starting in the background is automatically handled for us by the UP library, it sets up a dbus activatable service that launches FluffyChat with a new flag `--unifiedpush-bg`, which causes it to take the same codepath as Android. Additionally, the env var `FLUTTER_HEADLESS` is defined and handled on the GTK/C side, as otherwise a black window opens for a couple seconds before UP can trigger hiding it.

Likely some polish is needed, but it's a start, and a very important feature for those of us running FluffyChat on Linux mobile. I am happy to make any changes, please let me know :)

Fixes #486

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -77,6 +77,8 @@ dependencies:
   sqlcipher_flutter_libs: ^0.6.8
   swipe_to_action: ^0.3.0
   unifiedpush: ^6.2.0
+  unifiedpush_platform_interface: ^4.0.0
+  unifiedpush_storage_shared_preferences: ^1.0.0
   unifiedpush_ui: ^0.2.0
   universal_html: ^2.3.0
   url_launcher: ^6.3.2
--- a/integration_test/app_test.dart
+++ b/integration_test/app_test.dart
@@ -17,11 +17,170 @@ import 'users.dart';
 void main() {
   IntegrationTestWidgetsFlutterBinding.ensureInitialized();
 
-  group('Integration Test', () {
-    setUpAll(() {
-      // this random dialog popping up is super hard to cover in tests
-      SharedPreferences.setMockInitialValues({
-        'chat.fluffy.show_no_google': false,
+  group(
+    'Integration Test',
+    () {
+      setUpAll(
+        () async {
+          // this random dialog popping up is super hard to cover in tests
+          SharedPreferences.setMockInitialValues({
+            'chat.fluffy.show_no_google': false,
+          });
+        },
+      );
+
+      testWidgets(
+        'Start app, login and logout',
+        (WidgetTester tester) async {
+          app.main([]);
+          await tester.ensureAppStartedHomescreen();
+          await tester.ensureLoggedOut();
+        },
+      );
+
+      testWidgets(
+        'Login again',
+        (WidgetTester tester) async {
+          app.main([]);
+          await tester.ensureAppStartedHomescreen();
+        },
+      );
+
+      testWidgets(
+        'Start chat and send message',
+        (WidgetTester tester) async {
+          app.main([]);
+          await tester.ensureAppStartedHomescreen();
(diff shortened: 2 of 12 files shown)


TheSunCat:
I'm having some trouble testing it on my phone, mainly due to not being able to build an arm64 version from my x86_64 host. I'm looking into https://github.com/TheAppgineer/flatpak-flutter to make this easier (and just build it directly on my phone).

krille-chan:
Wow that is great! Thank you so much :-) can't wait to try this out

pubspec.yaml
```diff
@@ -25,25 +25,24 @@ dependencies:
...
   flutter_shortcuts_new: ^2.0.0
-  flutter_vodozemac: ^0.5.0
-  flutter_web_auth_2: ^5.0.1
-  flutter_webrtc: ^1.3.0
+  flutter_vodozemac: ^0.4.1
+  flutter_web_auth_2: ^4.1.0
+  flutter_webrtc: ^1.2.1
   geolocator: ^14.0.2
   go_router: ^17.0.1
-  handy_window: ^0.4.0
```
krille-chan: Is this package somehow producing problems here? That would be sad as it makes the app afaik easier integrated into linux mobile

TheSunCat: Yes, unfortunately its changes to GTK handling seem to conflict with running Flutter headless which is needed for background notifications.

Out of curiosity, do you know how `handy_window` affects Linux mobile? I am running a build with all my MRs merged on my own phone, and have not yet noticed any regressions.
I am running GNOME Mobile and cannot easily test other interfaces, so perhaps it's important for other UIs?

krille-chan: I only have heard it from other users but that's already years ago so maybe it is fixed now. I don't have a Linux mobile device for myself to test it. So if it works for you it is probably fixed. However it's still a little bit sad that this removes the rounded corners at the bottom of the window :-/ it looks so much more modern and it's not in sight that canonical develops a gtk4 port soon... However that would something we could life with I guess

TheSunCat: I'll try to see if I can find a way to make it work. I am currently struggling a lot with getting notifications in the background. I can start the app when a UP message comes through, but it never sends a notif.

TheSunCat:
I've been running a Flatpak built with all my open merge requests, and sometimes upon starting FluffyChat I get an error about the SQLite database being already in use, and the app opens to a black window. If I close and reopen it, I am shown the login screen and I have to re-login to my account (and verify my session). This sucks, and we cannot merge this until that's fixed. I would really appreciate some help!

pubspec.yaml
krille-chan: There was a new package of handy_window. Does that help in some case?

Codrax:
Any updates on this?

TheSunCat:
Well it works, I run all my patches on my phone, but it has the issue with the SQLite database causing me to have to relog every day or so. Still waiting for help with that.

PR #2596, opened 2026-02-20
feat: add autoSsoRedirect config option for automatic SSO login on web

swherdman:
## Summary

Adds an `autoSsoRedirect` config option for web deployments, similar to Element Web's [`sso_redirect_options.immediate`](https://github.com/element-hq/element-web/blob/develop/docs/config.md#sso_redirect_options). When enabled, unauthenticated users are automatically redirected to the homeserver's SSO provider on page load, bypassing the homeserver picker UI entirely.

This is useful for single-tenant Matrix deployments with a fixed homeserver and SSO-only login (e.g. Authentik, Keycloak), where the homeserver picker adds unnecessary friction.

### How it works
- Uses full-page redirect (`window.location.href`) to the SSO endpoint rather than flutter_web_auth_2's popup-based flow, which gets blocked by browsers without a user gesture
- `auth.html` stores the login token in `localStorage` and redirects back to the app
- On reload, the app recovers the token from `localStorage` and completes the Matrix login
- Router redirect skips the intro page and goes directly to the sign-in page when enabled

### Configuration
```json
{
    "defaultHomeserver": "chat.example.com",
    "autoSsoRedirect": true
}
```
Both `defaultHomeserver` and `autoSsoRedirect` must be set. Web-only; no effect on other platforms.

## Files changed
- `config.sample.json` — documents new option
- `lib/config/setting_keys.dart` — adds `autoSsoRedirect` setting
- `lib/config/routes.dart` — skips intro page when enabled
- `lib/widgets/view_model_builder.dart` — adds `onCreated` callback
- `lib/pages/sign_in/sign_in_page.dart` — wires `onCreated` to trigger auto-SSO
- `lib/pages/sign_in/view_model/sign_in_view_model.dart` — implements `autoSsoIfEnabled` flow
- `web/auth.html` — handles non-popup redirect mode via localStorage

## Test plan
- [ ] Set `autoSsoRedirect: true` + `defaultHomeserver` in config.json → unauthenticated user is redirected to SSO immediately
- [ ] After SSO login, app recovers token and lands on /rooms
- [ ] With `autoSsoRedirect: false` or unset, normal intro page flow is unchanged
- [ ] Non-SSO homeserver with `autoSsoRedirect: true` shows warning, falls back gracefully
- [ ] Non-web platforms are unaffected (guarded by `kIsWeb`)
- [ ] Existing popup-based SSO login (manual sign-in flow) still works

### DIFF
--- a/config.sample.json
+++ b/config.sample.json
@@ -25,5 +25,6 @@
   "noEncryptionWarningShown": false,
   "displayChatDetailsColumn": false,
   "colorSchemeSeedInt": 4283835834,
-  "enableSoftLogout": false
+  "enableSoftLogout": false,
+  "autoSsoRedirect": false
 }
--- a/lib/config/routes.dart
+++ b/lib/config/routes.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 import 'package:go_router/go_router.dart';
 import 'package:matrix/matrix.dart';
 
+import 'package:fluffychat/config/setting_keys.dart';
 import 'package:fluffychat/config/themes.dart';
 import 'package:fluffychat/pages/archive/archive.dart';
 import 'package:fluffychat/pages/bootstrap/bootstrap_dialog.dart';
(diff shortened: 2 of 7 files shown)


krille-chan:
Closed in favor of https://github.com/krille-chan/fluffychat/pull/2600

Issue #2686, opened 2026-03-11
Send Later/Schedule Send


skycatminepokie:
### Feature Description

An option to send a message later or to schedule a message to send later. Perhaps accessible by pressing and holding the send button.

### Rationale

Allows for sending messages while thinking about them, but allowing them to arrive at an acceptable time.

vedantpatel07756:
PR created for this enhancement  @krille-chan If needed further enhanced then let me know

skycatminepokie:
There's a matrix spec proposal that would make this a whole lot easier. Parts of the proposal are merged to synapse development branch. https://github.com/matrix-org/matrix-spec-proposals/pull/4140

PR #2750, opened 2026-03-22
Feat(Chat):Schedule Messages

vedantpatel07756:

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

**ISSUE :** Send Later/Schedule Send  #2686 

**Summary**
Add the ability to schedule messages to be sent at a specified date and time instead of sending them immediately.

**Changes**
New Schedule Button: Added a dedicated schedule button (clock icon) in the message input bar that appears only when text is entered

**Date/Time Picker:** Implemented date and time selection with proper formatting (12-hour format with AM/PM, no unnecessary zeros)

**Scheduled Messages Panel:** Added an expandable "Scheduled Messages" section above the input bar to display all pending scheduled messages with edit/delete capabilities

**Timer Management:** Messages are automatically sent at the scheduled time using Flutter's Timer API

**Cleanup:** Pending schedules are properly cancelled when the chat is closed

**UI/UX Improvements**
Schedule button visibility tied to input text state (hidden when empty)
Expandable panel for viewing scheduled messages without cluttering the input area
Individual delete buttons on each scheduled message to cancel scheduling
Clear date/time formatting using the app's existing localization system

Screenshot_1774204734

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -14,6 +14,7 @@ import 'package:fluffychat/pages/chat/event_info_dialog.dart';
 import 'package:fluffychat/pages/chat/start_poll_bottom_sheet.dart';
 import 'package:fluffychat/pages/chat_details/chat_details.dart';
 import 'package:fluffychat/utils/adaptive_bottom_sheet.dart';
+import 'package:fluffychat/utils/date_time_extension.dart';
 import 'package:fluffychat/utils/error_reporter.dart';
 import 'package:fluffychat/utils/file_selector.dart';
 import 'package:fluffychat/utils/matrix_sdk_extensions/event_extension.dart';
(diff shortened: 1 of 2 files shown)


lib/pages/chat/chat.dart
```diff
@@ -554,6 +569,12 @@ class ChatController extends State<ChatPageWithRoom>
     timeline = null;
     inputFocus.removeListener(_inputFocusListener);
     if (currentlyTyping) room.setTyping(false);
+
+    for (final scheduled in scheduledMessages) {
+      scheduled.timer.cancel();
+    }
+    scheduledMessages.clear();
```
krille-chan, line 576: So sheduled messages are canceled when leaving the chat page? They should be persistet and even then be executed when the app is not running anymore which is not yet possible on Web and Desktop btw and also hard on Android and iOS

Osthekake:
Is this a feature that someone wants? There are costs associated with a feature like this, and for people who have no use for scheduling messages, this will feature degrade their experience.

The area around the text input is prime real estate. How many random features can be supported there before it becomes cluttered and noisy?
Enabling/disabling things when user enters text will incur a small slow down. How many such slowdowns before the app becomes too slow?

PR #2773, opened 2026-03-28
Fix regression in Swedish translation

lochel:
This fixes a regression in the Swedish translation that I noticed with the recent update to version 1.5.1 of the android app.


krille-chan:
Hello, thank you very much but please only change translations over weblate so we have a sibgle source of truth. Otherwise we run into merge conflicts with weblate

lochel:
@krille-chan Thanks for your reply. I found a few more issues with the translation and will look into weblate to figure out how to address them. :+1:

PR #2775, opened 2026-03-28
feat: Add pasteboard

krille-chan:

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -35,6 +35,7 @@ import 'package:go_router/go_router.dart';
 import 'package:image_picker/image_picker.dart';
 import 'package:matrix/matrix.dart';
 import 'package:mime/mime.dart';
+import 'package:pasteboard/pasteboard.dart';
 import 'package:scroll_to_index/scroll_to_index.dart';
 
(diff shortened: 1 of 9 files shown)


lib/pages/chat/chat.dart
```diff
@@ -344,11 +345,39 @@ class ChatController extends State<ChatPageWithRoom>
       }
 
       return KeyEventResult.ignored;
+    } else if (evt is KeyDownEvent &&
+        evt.logicalKey == LogicalKeyboardKey.keyV &&
+        (HardwareKeyboard.instance.isControlPressed ||
+            HardwareKeyboard.instance.isMetaPressed)) {
+      _sendFileFromClipboard();
+      return KeyEventResult.handled;
```
Decodetalkers, line 353: we should return ignored to let the normal copy paste still work. because when the file is image, the it won't be paste to the input-row, so there is no worry

Decodetalkers:
It works on linux system.. and the library is still under good maintenance. it should work I think..

PR #2838, opened 2026-04-09
feat: unpin events directly from pinned events menu

ION606:
Replaces the action popup with a bottom sheet that shows unpin buttons next to each pinned event, which lets users to unpin without navigating to the event first.

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


krille-chan:
You can already unpin an event from the header. I would more target a UX where you can just tap on the header and then the visible pinned event changes inside of the header so you can toggle through it. Sorry, this does not fit into how I would like to have the UX

PR #2839, opened 2026-04-09
feat: right-click opens context menu on chat list items

ION606:
Adds right-click support to chat list items. For direct chats, shows the member actions popup menu. For group chats, falls back to the standard chat context action

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat_list/chat_list_body.dart
+++ b/lib/pages/chat_list/chat_list_body.dart
@@ -15,6 +15,7 @@ import 'package:matrix/matrix.dart';
 
 import '../../config/themes.dart';
 import '../../widgets/adaptive_dialogs/user_dialog.dart';
+import '../../widgets/member_actions_popup_menu_button.dart';
 import '../../widgets/matrix.dart';
 import 'chat_list_header.dart';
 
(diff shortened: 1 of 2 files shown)


lib/pages/chat_list/chat_list_body.dart
```diff
@@ -240,6 +241,19 @@ class ChatListViewBody extends StatelessWidget {
...
+                        final directChatMatrixId = room.directChatMatrixID;
+                        if (directChatMatrixId != null) {
+                          showMemberActionsPopupMenu(
+                            context: context,
+                            user: room.unsafeGetUserFromMemoryOrFallback(
+                              directChatMatrixId,
+                            ),
+                          );
+                        } else {
+                          controller.chatContextAction(room, context, space);
```
krille-chan, line 254: The right click here should do the same as onLongPress IMO

lib/pages/chat_list/chat_list_item.dart
```diff
@@ -69,7 +71,11 @@ class ChatListItem extends StatelessWidget {
         child: FutureBuilder(
           future: room.name.isEmpty ? room.loadHeroUsers() : null,
           builder: (context, _) => HoverBuilder(
-            builder: (context, listTileHovered) => ListTile(
+            builder: (context, listTileHovered) => GestureDetector(
+              onSecondaryTapUp: onSecondaryTap != null
```
krille-chan, line 75: I think this is the wrong callback. It should just be `onSecondaryTap`?

krille-chan (changes requested):
Adding right click support at one place would make the user think that right click support is everywhere where long press is also supported. So we should either add it everywhere or nowhere

lib/pages/chat_list/chat_list_body.dart
ION606: hmhm I think I was overthinking it :/

ION606:
> Adding right click support at one place would make the user think that right click support is everywhere where long press is also supported. So we should either add it everywhere or nowhere

I mean I suppose, I'll work on it when job allows ig

PR #2840, opened 2026-04-09
feat: support pasting images from clipboard with Ctrl+V

ION606:
Adds pasteboard dependency and handles Ctrl+V (or Cmd+V) to paste images directly from the clipboard on desktop. Falls back to standard text paste when no image is found.

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### A Note!

This adds the [pasteboard](https://pub.dev/packages/pasteboard) package dependancy!


krille-chan (changes requested):
This does not work for me :-( I've tested it locally on my Ubuntu 24.04 machine in debug mode and nothing happens for files. This only seem to support images. However Pasteboard.files() exists?

I tested also with an image but this only pastes the file name into the textfield.

However when I change it to Pasteboard.files() it works very nice for all kind of files. Can you change it to something like this?

```dart
final files = await Pasteboard.files();
    if (files.isNotEmpty) {
      if (!mounted) return;
      showAdaptiveDialog(
        context: context,
        builder: (c) => SendFileDialog(
          files: files.map(XFile.new).toList(),
          room: room,
          outerContext: context,
          threadRootEventId: activeThreadId,
          threadLastEventId: threadLastEventId,
        ),
      );
      return;
    }
// other stuff
```

krille-chan:
@ION606 are you still going to work on this?

Orhideous:
Tested locally, works. **Finally!** :)

krille-chan (approved):

PR #2842, opened 2026-04-09
feat: show sticker body text in chat list preview

ION606:
Displays the sticker name/body in the chat list subtitle instead of a generic message, so users can see which sticker was sent. I also changed the emoji to the...closest thing I could find to a sticky note (please let me know if this wasn't a good choice).

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -1673,7 +1673,7 @@
             }
         }
     },
-    "sentAStickerBody": "🗒️ {username} sent {stickerBody}",
+    "sentAStickerBody": "🗒️ {username} sent sticker {stickerBody}",
     "@sentAStickerBody": {
         "type": "String",
         "placeholders": {
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -1673,6 +1673,18 @@
             }
         }
     },
+    "sentAStickerBody": "🗒️ {username} sent {stickerBody}",
+    "@sentAStickerBody": {
+        "type": "String",
+        "placeholders": {
+            "username": {
+                "type": "String"
+            },
+            "stickerBody": {
+                "type": "String"
+            }
+        }
+    },
     "sentAVideo": "🎥 {username} sent a video",
     "@sentAVideo": {
         "type": "String",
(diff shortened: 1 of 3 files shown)


lib/pages/chat_list/chat_list_item.dart
```diff
@@ -323,7 +336,18 @@ class ChatListItem extends StatelessWidget {
...
+                                    lastEvent!
+                                        .senderFromMemoryOrFallback
+                                        .calcDisplayname(
+                                          i18n: MatrixLocals(
+                                            L10n.of(context),
+                                          ),
+                                        ),
+                                    lastEvent.body,
+                                  )
+                                : lastEvent?.calcLocalizedBodyFallback(
```
krille-chan: This is the wrong place to do it. Please instead use the `MatrixLocalization` class

lib/l10n/intl_en.arb
```diff
@@ -1673,6 +1673,18 @@
             }
         }
     },
+    "sentAStickerBody": "🗒️ {username} sent {stickerBody}",
```
krille-chan: The stickerBody is generated by a user. So a user could misuse this like setting the stickerBody to "a video" which would result int "Username sent a video" while the message type is actually different. Not sure if this is a good way to do

ION606: hmhm I thought so too, so I put the sticker emoji, but I can add "sticker" to it :3

lib/utils/matrix_sdk_extensions/matrix_locals.dart
```diff
@@ -5,7 +5,26 @@ import 'package:matrix/matrix.dart';
...
+      final senderName =
+          event.senderFromMemoryOrFallback.calcDisplayname(i18n: i18n);
+      if (i18n is MatrixLocals && body.isNotEmpty) {
+        return i18n.l10n.sentAStickerBody(senderName, body);
+      }
+      return i18n.sentASticker(senderName);
+    };
+  }
+
+  MatrixLocals(this.l10n) {
```
krille-chan, line 25: Hm, this is still an unnecessary complex workaround. I see you need the body from the event so you cannot just overwrite `String sentASticker(String senderName)`. However, this should then be patched in the Matrix Dart SDK instead of doing it here so that `sentASticker` also gives out the body for optional use.

Issue #1329, opened 2024-09-11
Proposal: Improve text rendering on macOS and iOS


reinhart1010:
Hi, I'd like to make another PR to improve how FluffyChat looks on iOS. Historically, FluffyChat was subject to a 4-year-old Flutter text rendering bug (https://github.com/flutter/flutter/issues/43998) that affects devices running iOS 13 and later. And it impacted most Flutter apps and widget components which are specifically based on `MaterialApp`.

The issue has been fixed as of February 2024, so the SF Pro Display font (as `CupertinoSystemDisplay`) should now be used in place of SF Pro Text (as `CupertinoSystemText`) for text headings, assuming that the app is built using the latest version of Flutter SDK. And I can already see this new behavior in FluffyChat.

![image](https://github.com/user-attachments/assets/98442099-41cc-4f3b-8e41-37a4e67f6f6f)

Now, to finally fix the bug where text on FluffyChat doesn't look as tidy as other apps, the `letter-spacing` values of the app's `TextTheme` must be adjusted to render the Apple's font as what Apple has intended, not Google's. tl;dr, as I've [researched](https://reinhart1010.id/blog/2023/02/11/why-text-labels-on-flutters-materialapp-looks-worse-on-ios-and-macos) before, this is why 

| Style | Font Size | Material Design's `letter-spacing` (Actual) | Apple HIG's `letter-spacing` (Expected) | Difference |
|---|---|---|---|---|
| Body Large | 16 | 0.5 | -0.31 | **0.81** |
| Body Medium | 14 | 0.25 | -0.15 | **0.4** |
| Body Small | 12 | 0.4 | 0 | **0.4** |

bitrage:
Please make this happen, looks oddly out of place

PR #2891, opened 2026-04-24
Match system typography on macOS

borzov:
Hi! I've been using FluffyChat on macOS for a while and the thing that kept bothering me is how the text never quite fits in with the rest of the desktop. Next to Mail, Messages or Telegram Desktop the fonts are visibly smaller and a little "off" — wrong family, too much letter-spacing, too much vertical space between lines. Once you notice it you can't un-notice it, and there was no slider or setting that fixed it cleanly.

This PR is a coordinated attempt to close that gap on macOS while not touching how the app looks on the other platforms.

## What changes

**The font-size slider now affects the whole UI, not just bubbles.** Previously `AppSettings.fontSizeFactor` was multiplied into ~28 call sites inside message widgets, so moving the slider in Settings → Style would only resize message text; chat list, app bar, dialogs and the compose input stayed put. I wrapped the router in a MediaQuery that combines the OS `textScaler` with the user factor and deleted those 28 manual multiplications. Now the slider affects everything, and iOS Dynamic Type / Android font scale still work — they compose multiplicatively with the user's preference.

**SF Pro on Apple platforms.** The theme now sets `fontFamily: 'CupertinoSystemText'` on iOS and macOS, which the Flutter engine resolves to `.AppleSystemUIFont` with the native Display↔Text split at 20pt. The TextTheme also gets Apple HIG letter-spacing for body / label / title roles (iOS body uses `−0.41`, macOS uses near-zero or slightly positive), so SF Pro is rendered the way Apple intended instead of being run through Material's `+0.25` / `+0.5` tracking that was optimised for Roboto. The composer TextField picks up the same size so typing matches the bubbles.

**Message body is 14pt on macOS (still 16pt elsewhere).** `AppConfig.messageFontSize` is now platform-aware. 14pt is what Telegram Desktop, Mail and the other native macOS chat apps use; 16pt was Material's Android-centric default and felt out of place on the desktop.

**Tighter line-height on macOS.** A new `AppConfig.messageLineHeight = 1.25` is threaded through every TextStyle that renders body-sized text — plain text, Linkify styles, reply quotes, image / video / file captions, poll question + answers, HtmlMessage paragraphs and blockquotes, audio player caption and the Settings → Style preview. Off-macOS this value is \`null\`, so every TextStyle falls back to the font's intrinsic leading and nothing visually changes. HtmlMessage headings (h1–h6) keep their own higher \`height\` values because at larger sizes Material-style breathing room still reads better than a compressed 1.25.

**Chat list / app bar hierarchy.** As a side-effect of the slider fix, the gap between sender title and secondary text (preview, timestamp, presence) was uncomfortably wide — roughly −4 to −5pt instead of Telegram's −1 to −3. I lifted six hardcoded \`fontSize: 11\` / \`fontSize: 12\` values by one point across \`chat_list_item\`, \`chat_app_bar_title\`, \`message.dart\` (sender name / bubble time / read marker) and \`state_message\`. Small change; big readability win.

## Why one PR

The pieces are interdependent. The global MediaQuery wrapper would double-scale every bubble without the manual-multiplier cleanup. The 14pt macOS body only makes sense once \`platformFontFactor\` isn't multiplying everything else on top. The line-height constant wouldn't land without the AppConfig refactor. So it's 19 files in one signed commit (+218 / −80, one new file — \`lib/config/app_font_size.dart\` — for the \`ValueNotifier<double>\` and the \`ScaledTextScaler\` that the MediaQuery wrapper uses). Happy to split it if you'd prefer, but each concern depends on the previous one landing.

## Verification

- \`dart format lib\` — clean.
- \`flutter analyze\` — no issues.
- \`flutter test\` — 4/4 passing.
- Manually tested on macOS with the slider at 0.5, 1.0, 1.5 and 2.5; body, chat list, settings, dialogs and compose all respond. Default slider (1.0) on a Retina display now looks like what you'd expect from a native app sitting next to Mail or Telegram Desktop.
- On Android / iOS / Linux / Windows / Web the visible diff is a \`+1pt\` bump on six hardcoded secondary-text sites and nothing else — \`messageLineHeight\` is \`null\` off-macOS, \`messageFontSize\` stays at 16pt, and the Apple \`fontFamily\` / letter-spacing branches only fire on iOS and macOS.

## Related

Refs #1329. This doesn't touch the iOS Dynamic Type letter-spacing table from that proposal directly, but it does ship the \`fontFamily\` + letter-spacing side and adds a macOS-specific answer alongside it.

### DIFF
--- a/lib/config/app_config.dart
+++ b/lib/config/app_config.dart
@@ -1,4 +1,7 @@
+import 'dart:io' show Platform;
 import 'dart:ui';
+
+import 'package:flutter/foundation.dart';
 
 abstract class AppConfig {
   // Const and final configuration values (immutable)
@@ -7,7 +10,20 @@ abstract class AppConfig {
   static const Color secondaryColor = Color(0xFF41a2bc);
 
   static const Color chatColor = primaryColor;
-  static const double messageFontSize = 16.0;
+  // Chat body base size. macOS native chat apps (Telegram, Messages, Mail)
+  // render body around 14pt, noticeably tighter than Material 3's 16sp.
+  // Everywhere else we keep Material/HIG defaults (16sp on Android,
+  // ~17pt on iOS via Dynamic Type).
+  static final double messageFontSize = (!kIsWeb && Platform.isMacOS)
+      ? 14.0
+      : 16.0;
+
+  // Line-height for message body. macOS/SF Pro feels natural at 1.25×,
+  // which is tighter than the default Material/font leading. On other
+  // platforms we leave it null so each font uses its own intrinsic leading.
+  static final double? messageLineHeight = (!kIsWeb && Platform.isMacOS)
+      ? 1.25
+      : null;
   static const bool allowOtherHomeservers = true;
   static const bool enableRegistration = true;
   static const bool hideTypingUsernames = false;
--- a/lib/config/app_font_size.dart
+++ b/lib/config/app_font_size.dart
@@ -1,0 +1,36 @@
+import 'package:flutter/widgets.dart';
+
+import 'setting_keys.dart';
+
+/// Listenable mirror of [AppSettings.fontSizeFactor] so the global
+/// [MediaQuery.textScaler] wrapper rebuilds when the slider moves.
+final ValueNotifier<double> fontSizeFactorNotifier = ValueNotifier<double>(
+  AppSettings.fontSizeFactor.value,
+);
+
+/// Multiplies an inner [TextScaler] by a linear factor, preserving
+/// non-linear OS behavior (e.g. iOS Dynamic Type) on top of the user
+/// font-size preference.
+@immutable
+class ScaledTextScaler extends TextScaler {
+  const ScaledTextScaler(this.inner, this.factor);
+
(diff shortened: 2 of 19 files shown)


krille-chan:
Is this ai slop?

borzov:
> Is this ai slop?

Fair - yeah, I used AI to write the description. My English isn't strong enough to get something that long readable on my own. The problem and the code are mine though... I've been looking at FluffyChat next to Mail and Telegram Desktop on my Mac for a while and the text really does stand out ((( If the long writeup is the issue, I can strip it down to a one-paragraph summary in my own words, or split the change into smaller PRs where each one speaks for itself.

borzov:
I'm using this build on my computer, and if needed, I can show you what the message looks like after applying these changes. I can even compare it to the original version.

krille-chan:
I doubt that you wrote this on your own. This looks way too complex for a problem which probably can be fixed by removing stuff instead of adding it, so I assume the code is also AI slop. I don't wanna go away from Flutter defaults anyway.

PR #2137, opened 2025-08-19
feat: add GIF picker integration with Tenor API

ariqpradipa:
*Thank you so much for your contribution to FluffyChat ❤️❤️❤️*

Please make sure that your Pull Request meet the following **acceptance criteria**:

- [x] Code formatting and import sorting has been done with `dart format lib/ test/` and `dart run import_sorter:main --no-comments`
- [x] The commit message uses the format of [Conventional Commits](https://www.conventionalcommits.org)
- [x] The commit message describes what has been changed, why it has been changed and how it has been changed
- [] Every new feature or change of the design/GUI is linked to an approved design proposal in an issue
- [ ] Every new feature in the app or the build system has a strategy how this will be tested and maintained from now on for every release, e.g. a volunteer who takes over maintainership

### Pull Request has been tested on:

- [ ] Android
- [x] iOS
- [x] Browser (Chromium based)
- [x] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

address #700 

Screenshot 2025-08-19 at 6 46 43 AM
Screenshot 2025-08-19 at 6 46 56 AM

Tenor requires an API key. Currently, I’ve set the key to `<tenor-api-key>`.
If this somehow gets merged, it should be replaced with a real API key.

I haven’t found any pricing information for the Tenor API. However, to increase the request rate (beyond the default limit of 1 request per second), we would need to contact the Tenor team.

Issue #700, opened 2023-11-28
[Feature Request] Tenor GIFs Search


Heus-Sueh:
### Feature Description

Implement [Tenor GIFs](https://tenor.com/) with a search engine
Commet's dev managed to implement it in his app somehow, perhaps it would be interesting to see how the implementation was done to implement it in Fluffy
https://github.com/commetchat/commet/tree/main/commet/lib/client/components/gif

### Rationale

GIFs are a fun way to communicate with other people and this feature is present in "any" communication app

### Mockup

the design could be similar or the same as telegram:
![image](https://github.com/krille-chan/fluffychat/assets/98482469/5c94ef00-64b5-439f-8378-f7d5f66124e0)

Mobile:
![image](https://github.com/krille-chan/fluffychat/assets/98482469/cab799d7-264b-4b54-968b-c02faf68bf5a)

### Additional Context

none

MaximilianZenz:
+1 An integrated GIF search would be great.

nebulade:
Just for kicks, if fluffy gets this faster than element it would be a joke on their end https://github.com/element-hq/element-meta/issues/321 :D

ariqpradipa:
I think one of the main challenges is how we use the Tenor API key. By default, a newly created key is limited to one request per second, unless we reach out to the Tenor team to have this limit increased. Right now, I’m experimenting with implementing GIFs using my own API key, but for production use, we’ll likely need a different solution.

PR #2938, opened 2026-05-12
feat: add gif picker and favorite gifs

JuLi0n21:

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

Implements #700 
Tenor is dead, this is a replacement build using Giphy as the api
the code builds on top #https://github.com/krille-chan/fluffychat/pull/2137 and the changed file sending behavior from https://github.com/ItsKaa/fluffychat/commit/3ca6d9c79793621f028ed0afaeaa89401f663343 

now because giphy seems to be very pricey, a place in the chat / emoji configuration was added to input ur own apikey with a mini guide, this should be replaced if we use a commercial lisence (i have not read the tos if this is allowed) secondly caching infrasturce will defently be required to cut costs, currently the gif ui is hidden if no apikey is provided.

currently the [rating](https://support.giphy.com/hc/en-us/articles/360058840971-Content-Rating) is hardcoded to `r` if this should be editable/ changed pls say so

i was also looking for ideas to have favorite gifs in the same tab, wasnt sure how to best handle the persistent storage required

Screenshots:
image
image


krille-chan:
I really appreciate your work and contribution. Unfortunately I'm concerned about the privacy impact of embedding giphy into our product. Yes it could be optional. However also the solution with the token is something I'm concerned because adding your own token is very complicated for normal users so this would only be useful for a very tiny fraction of users while it seems to be a quite complex feature which could easily break. According to the design philosophy of FluffyChat to keep everything very minimalistic (and therefore easy to maintain) I'm not sure that this fits into it. Also because there are several workarounds. Most keyboards on Android and the default iOS keyboard already have giphy embedded into it and it is quite easy to go to the giphy website and copy-paste them. So this is only a convenient feature for a very small fraction of the users.

Maybe a compromise: We make pasting images possible instead so for desktop users it becomes much easier to copy-paste gifs from the giphy website: https://github.com/krille-chan/fluffychat/pull/2840

TheAnnoyingFruit:
Klipy is another option for gif api

uvulpos:
> I really appreciate your work and contribution. Unfortunately I'm concerned about the privacy impact of embedding giphy into our product. Yes it could be optional. However also the solution with the token is something I'm concerned because adding your own token is very complicated for normal users so this would only be useful for a very tiny fraction of users while it seems to be a quite complex feature which could easily break. According to the design philosophy of FluffyChat to keep everything very minimalistic (and therefore easy to maintain) I'm not sure that this fits into it. Also because there are several workarounds. Most keyboards on Android and the default iOS keyboard already have giphy embedded into it and it is quite easy to go to the giphy website and copy-paste them. So this is only a convenient feature for a very small fraction of the users.
> 
> 
> 
> Maybe a compromise: We make pasting images possible instead so for desktop users it becomes much easier to copy-paste gifs from the giphy website: https://github.com/krille-chan/fluffychat/pull/2840

I share your concern, however we could add a compromise here by implementing a gif hub. Like bring your own gifs / memes. And then you can enhance it yourself with external sources such as giphy or whatever you use. And you can import posted images to your library. It think that's a good compromise. I really want gifs in fluffychat

JuLi0n21:
@krille-chan Hi, the correct way to add kilpy or giphy would be to hardcode a pair of keys and run all user requests over that, i didnt want to do that yet, aswell as that it will cost money, in addition its recommended to setup some custome infrastructe to reduce requests to a minimum using chacing and cds

in addition theirs a favorite gif tab now, automatic embedding was added with an setting to turn it on or off, since im not using many clients i can say how well the current approach works with the other exstiting clients
some new screenshots will follow in a bit

JuLi0n21:
Screenshot_20260724-164721_FluffyChat
Screenshot_20260724-164617_FluffyChat
Screenshot_20260724-164605_FluffyChat
Screenshot_20260724-164539_FluffyChat
Screenshot_20260724-164528_FluffyChat
Screenshot_20260724-164510_FluffyChat

Issue #2820, opened 2026-04-04
Increase or make configurable message preview height for long [bot] messages


kaskavalci:
### Summary

Long plain or HTML `m.room.message` texts are shown with a fixed preview height when the message is not selected: `HtmlMessage` uses `maxLines: 64` and `TextOverflow.fade` (`lib/pages/chat/events/html_message.dart`). This is independent of `textMessageMaxLength`, which only affects the composer.

For **bot and automation** use cases, messages are sometimes long (logs, reports, formatted status). Users see only the first portion with a fade until they select the bubble, which is easy to miss and hurts readability for non-interactive "wall of text" bots.

### Current behavior

- Unselected: content is capped at **64 lines**, then fades.
- Selected: `limitHeight` is false → `maxLines` is unbounded.  Everything is visible.

### Desired behavior

One or more of:

1. **Configurable** max preview lines (e.g. `AppSettings` + Advanced configs / `config.json`, consistent with other tuning knobs), or
2. A **higher default** for the line cap, or
3. Another UX that makes long bot messages readable without requiring users to discover "tap to expand" (documented tradeoffs welcome).

### Optional notes

- Platform (observed): iOS

Thanks for considering this. Happy to chime in for its implementation if you agree.

### Rationale

- Fluffychat is an amazing platform for matrix, it's a shame that long texts are hidden. This is not an issue on elements app, for instance.
- For long messages, especially from bots, it's crucial that user sees the complete message.

krille-chan:
Sounds reasonable. Maybe make multiple things? Make it configurable **and** increase the default length from 64 lines to 128? That should still protect from spammers with messages with 3000 lines length to break rooms

kaskavalci:
Great that it's accepted. I'll work on it on the first chance. I agree to make it configurable. Are you OK with the advanced configs view?

kaskavalci:
Closes this issue: https://github.com/krille-chan/fluffychat/pull/2847

gokussjx:
I have a slightly different take on this, allowing a tap to "show more" and another to "show less" for any message that exceeds the cutoff. Created a PR #3000 , but happy to discuss more on the design if there's any question/suggestion. I also tested another option akin to iMessage where long messages can open up a full page modal view, but opted against it since it takes you away from the "thread" of conversation and hides all other messages in context.

PR #3000, opened 2026-05-24
feat: collapsible long messages with show more/less

gokussjx:
Long messages are clipped with `TextOverflow.fade` at 128 lines with no way to read the full content except press-hold selecting the bubble. This replaces the fade with an inline "Show more" / "Show less" toggle.

- Reduce default `messagePreviewMaxLines` from 128 to 20
- Use a DOM text heuristic (line count + character length) to detect messages that will overflow, since `TextPainter` cannot measure `WidgetSpan`s
- Add `_CollapsibleText` widget with `AnimatedSize` for smooth expand/collapse
- Add L10n keys `showMore` and `showLess`

Short messages render without any button. Long messages show truncated with ellipsis and a "↓ Show more" link. Tapping expands the full message with a smooth 200ms animation and the link changes to "↑ Show less" to collapse back.

Refs #2820

---

- [x] Android
- [ ] iOS
- [ ] Browser (Chromium based)
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

share_6759720282158837964

Screenshot_20260524-160112

### DIFF
--- a/lib/config/setting_keys.dart
+++ b/lib/config/setting_keys.dart
@@ -16,7 +16,7 @@ enum AppSettings<T> {
   textMessageMaxLength<int>('textMessageMaxLength', 16384),
 
   /// Max lines for unselected HTML/text bubbles; 0 = unlimited (no fade).
-  messagePreviewMaxLines<int>('chat.fluffy.message_preview_max_lines', 128),
+  messagePreviewMaxLines<int>('chat.fluffy.message_preview_max_lines', 20),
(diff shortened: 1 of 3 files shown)


krille-chan (approved):
The code looks very clean and I don't think this should affect the performance too much. The heuristic to check if the text "likely overflows" also looks reasonable to me! Thank you very much, this was really nice to review! :-)

Issue #2857, opened 2026-04-13
Support for MSC4028


shibotto:
### Feature Description

It would be nice for FluffyChat to handle the proposed MSC4028 (Push all encrypted events except for muted rooms).

### Rationale

[MSC4028](https://github.com/matrix-org/matrix-spec-proposals/blob/giomfo/push_encrypted_events/proposals/4028-push-all-encrypted-events-except-for-muted-rooms.md) pushes all encrypted events to clients. The purpose is for push enabled clients to have mentions-and-keywords-only notifications in encrypted rooms as well. This is already implemented in Synapse (experimental flag) with rule_id `.org.matrix.msc4028.encrypted_event` and Element X when choosing "Mentions and Keywords only".

From the proposal:

> When this push rule matches then homeserver would push the event to all registered clients, similar to how other rules work. Clients would decrypt the event and re-run push rules as normal. If the decrypted event results in no notification, the push is discarded. If the decryption failed, the new .m.rule.encrypted_event rule will match on client side, but there is no reason to notify the end user, so the push is discarded too. Globally the client must discard the push when the event stays encrypted locally because this use case should not happen except if the event was not encrypted for the current session, or if we are in front of an unexpected "Unable To Decrypt" (UTD) issue.

As of now I think FluffyChat never discards these events, which results in a notification for every encrypted event.

krille-chan:
A clear NO from my side! This is destroying the battery and a super bad idea. We are waitong for a flat in the unencrypted content which indicates that a message contains a mention or other high prio content. There was also a msc for it iirc

shibotto:
I'm not really a fan of leaking more context about encrypted conversations in clear, but do you by any chance remember which msc you are referring to? I'd like to keep an eye on that one anyway, because right now having to check Fluffy every minute for new messages is what's destroying my battery.

PR #3010, opened 2026-05-25
feat: Client-side push notification filtering for encrypted rooms

PolynomialDivision:
MSC3952 (Intentional Mentions) was rewritten to no longer allow signalling mentions in the unencrypted wrapper of encrypted events, as it would leak metadata about room activity to the server.

Because of this, users who want to receive mention or @room notifications in encrypted rooms have to enable push notifications for all encrypted messages. FluffyChat did not filter these client-side since the device wakes up for every message regardless, meaning battery drains the same whether notifications are shown or not. However, this causes the device to be spammed with a notification for every single message.

Adds an opt-in setting under the encrypted room notification push rule that decrypts each incoming message on arrival and evaluates the full push rule set locally, replicating the same filters the user has configured for unencrypted rooms. This stops the spam without any additional battery impact.
1:1 encrypted rooms are unaffected as push rules already notify for all direct messages.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/utils/push_helper.dart
+++ b/lib/utils/push_helper.dart
@@ -160,6 +160,11 @@ Future<void> _tryPushHelper(
   }
   Logs().v('Push helper got notification event of type ${event.type}.');
 
+  if (!client.pushruleEvaluator.match(event).notify) {
+    Logs().i('Push helper: filtered by client-side push rules.');
+    return;
+  }
+
   if (event.type.startsWith('m.call')) {
     // make sure bg sync is on (needed to update hold, unhold events)
     // prevent over write from app life cycle change


PolynomialDivision:
I did not create an issue, I hope this is fine.

PolynomialDivision:
friendly reminder @krille-chan : Maybe you can start the pipeline (workflows) so I can see if the code changes have already some issues?

lib/utils/push_helper.dart
```diff
@@ -160,6 +160,14 @@ Future<void> _tryPushHelper(
   }
   Logs().v('Push helper got notification event of type ${event.type}.');
 
+  if (event.room.encrypted &&
+      AppSettings.clientSidePushFilterEncrypted.value &&
+      event.type != EventTypes.Encrypted &&
+      !client.pushruleEvaluator.match(event).notify) {
```
krille-chan: The method `pushruleEvaluator` is actually very safe. We can use it all the time without all the preconditions or the feature flag and only benefit from it I think. No need to make it optional. While I would increase the log level:
```suggestion
      if (!client.pushruleEvaluator.match(event).notify) {
          Logs().i('Push helper: filtered by client-side push rules.');
          return;
      }
```

This would simplify the whole PR as we don't need the feature flag

PolynomialDivision: I've removed the feature flag and all preconditions, so the push rule evaluator now runs unconditionally for every event. Also bumped the log level to .i as you suggested.

krille-chan (approved):

lib/utils/push_helper.dart
krille-chan: Thank you so much :)

shibotto:
Thank you for finally supporting [MSC4028](https://github.com/krille-chan/fluffychat/issues/2857) ❤️

PR #3046, opened 2026-05-30
chore: remove webcrypto override

### LINKED ISSUE #207: Could provide a version adapted for Android 15 with a 16KB size limit?
problem: Starting November 1, 2025, all new apps and updates to existing apps submitted to Google Play that target Android 15 and higher must support a 16 KB page size on 64-bit devices.

https://developer.android.com/guide/practices/page-sizes?hl=zh-cn

twoneis:
https://github.com/google/webcrypto.dart/pull/247 closes https://github.com/google/webcrypto.dart/issues/207 which was the reason for the pin and is included in 0.6.1. 

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -5,26 +5,26 @@ packages:
     dependency: transitive
     description:
       name: _fe_analyzer_shared
-      sha256: "1dd467c7e56541bea70bbd35d537e3aa12dfba81f39e2a75bb6a61fc5595985b"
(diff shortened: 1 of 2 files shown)


krille-chan:
@twoneis thanks but I think you ran `flutter pub get` not with the correct flutter version so the CI blocks you here

twoneis:
@krille-chan thanks, fixed that and also noticed it was only a transitive dependency before which caused another CI check to fail but I think it's good now.

PR #3050, opened 2026-05-31
feat: add inline PDF viewer for Matrix file attachments

### LINKED ISSUE #1477: Enhancing with Integrated PDF Viewing
### Feature Description

It would be a fantastic addition to Fluffychat's user-friendly interface to allow PDFs to be viewed directly within the client. When a PDF document is clicked, it should open seamlessly inside Fluffychat. This could be achieved using the following package:

https://pub.dev/packages/flutter_pdfview

### Rationale

Adding the ability to view PDFs directly within Fluffychat would enhance the user experience by providing a more seamless and integrated interaction.

### Mockup

_No response_

### Additional Context

_No response_

PolynomialDivision:
I built this feature with the help of Claude Code. I couldn't find any guidelines against AI-assisted contributions, so I hope that's fine. I've been missing inline PDF viewing in FluffyChat for a while and wanted to contribute it, even though I'm not a Dart/Flutter developer by trade. Happy to iterate on any feedback.

Web inline PDF rendering is currently disabled: I wasn't able to test it properly. The pdfx web plugin is already included and the platform guard in DefaultAttachmentRenderer can be removed to enable it once pdf.js is set up and tested.

----

Introduces a MIME-aware attachment rendering layer and a dedicated PDF preview pipeline using pdfx. PDFs render inline in the chat bubble (tap to load, tap again for fullscreen pinch-zoom viewer) with a 50 MB fallback to the external OS viewer.

Verified on Android. iOS uses the same native renderer path. Linux and Web fall back to the external OS viewer.

Web PDF rendering is intentionally disabled for now. The pdfx web plugin is included in the build but not activated. Web inline PDF can be enabled in a follow-up by adding pdf.js and removing the web guard in DefaultAttachmentRenderer.

Also includes:
- First-page thumbnail cached to disk, shown in idle state on revisit
- AutomaticKeepAliveClientMixin to preserve loaded state across scrolls
- NeverScrollableScrollPhysics to prevent nested scroll conflicts
- Re-entry guard against concurrent download calls
- RepaintBoundary to isolate PDF rasterisation from chat list repaints
- PdfCache stale-path guard for iOS temp-dir purges
- Page resource always released via finally block in thumbnail generator
- cacheWidth hint on thumbnail Image to decode at display size
- Loading spinner in fullscreen viewer while document initialises

Fixes: #1477

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desk
- [ ] Desktop Windows
- [ ] Desktop macOS

photo_4_2026-05-31_11-14-45
photo_3_2026-05-31_11-14-45
photo_2_2026-05-31_11-14-45
photo_1_2026-05-31_11-14-45

### DIFF
--- a/lib/attachment/attachment_renderer.dart
+++ b/lib/attachment/attachment_renderer.dart
@@ -1,0 +1,14 @@
+// SPDX-FileCopyrightText: 2019-Present Christian Kußowski
+// SPDX-FileCopyrightText: 2019-Present Contributors to FluffyChat
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import 'package:flutter/material.dart';
+import 'package:matrix/matrix.dart';
+
+import 'attachment_style.dart';
+
+abstract class AttachmentRenderer {
+  const AttachmentRenderer();
+  Widget build(BuildContext context, Event event, AttachmentStyle style);
+}
--- a/lib/attachment/attachment_style.dart
+++ b/lib/attachment/attachment_style.dart
@@ -1,0 +1,18 @@
+// SPDX-FileCopyrightText: 2019-Present Christian Kußowski
+// SPDX-FileCopyrightText: 2019-Present Contributors to FluffyChat
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import 'package:flutter/material.dart';
+
+class AttachmentStyle {
+  final Color textColor;
+  final Color linkColor;
+  final BorderRadius borderRadius;
+
+  const AttachmentStyle({
+    required this.textColor,
+    required this.linkColor,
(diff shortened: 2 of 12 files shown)


krille-chan:
Thanks for the reminder to add guidelines against AI slop contributions ❤️

krille-chan:
So I don't aggree that we should add the maintenance burden of an in-app-pdf-viewer into FluffyChat. I recentely added an "Open" button to the download-success-snackbar which makes it super easy to download and directly open an pdf in the default PDF viewer of the OS. The UX is nearly the same without adding such a big package which makes maintenance work in the long time on my other projects already. So I'll close this

PR #3052, opened 2026-05-31
feat: configurable message collapse threshold in chat settings

jonharris0n:
## What

Adds a user-configurable threshold for collapsing long messages in the chat
timeline, surfaced as a slider in **Settings → Chat**.

## Why

Long plain-text / HTML messages are collapsed at a fixed line count today
(see the `messagePreviewMaxLines` behavior in `HtmlMessage`). Different users
want different limits — some want aggressive collapsing to keep the timeline
scannable, others want to see more inline before the "show more" fold. This
exposes the existing setting instead of leaving it hardcoded.

## How

- Surfaces the existing `AppSettings.messagePreviewMaxLines` setting
  (key `chat.fluffy.message_preview_max_lines`, default `25`) via a
  `Slider.adaptive` in the chat settings page.
- Slider range: floor derived from `AppSettings.messagePreviewMaxLines.defaultValue`
  (single source of truth, no duplicated literal), max `100`.
- A value of `0` disables collapsing entirely — `HtmlMessage` already treats
  `<= 0` as "never collapse" (`maxLines == null`).

## Changes

| File | Change |
|------|--------|
| `lib/l10n/intl_en.arb` | New strings: `messagePreviewMaxLines`, `messagePreviewMaxLinesDescription` |
| `lib/pages/settings_chat/settings_chat.dart` | `changeMessagePreviewMaxLines(double)` setter |
| `lib/pages/settings_chat/settings_chat_view.dart` | Slider UI bound to the setting |

1 commit, 3 files, +38 lines.

## Testing

- `dart analyze` clean on all changed files plus the render-path consumer
  (`lib/pages/chat/events/html_message.dart`).
- macOS profile build succeeds; app launches without errors.
- Verified the value pipeline end-to-end by inspection: setting key → controller
  setter → slider UI → render-time consumer at `html_message.dart` reads
  `AppSettings.messagePreviewMaxLines.value` and gates collapse on it, with the
  `<= 0 → no collapse` guard intact.

## Notes

- Collapse triggering uses the existing heuristic in `HtmlMessage`
  (`lineCount > maxLines || plainText.length > maxLines * 50`), not a true
  rendered-line measurement. This PR only makes the threshold configurable;
  it does not change the detection strategy. Happy to revisit that separately
  if maintainers prefer a measured approach.

### DIFF
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -2335,6 +2335,8 @@
     "sendReadReceiptsDescription": "Other participants in a chat can see when you have read a message.",
     "formattedMessages": "Formatted messages",
     "formattedMessagesDescription": "Display rich message content like bold text using markdown.",
+    "messagePreviewMaxLines": "Collapse long messages",
+    "messagePreviewMaxLinesDescription": "Messages longer than this many lines collapse behind a Show more button.",
     "verifyOtherUser": "🔐 Verify other user",
     "verifyOtherUserDescription": "If you verify another user, you can be sure that you know who you are really writing to. 💪\n\nWhen you start a verification, you and the other user will see a popup in the app. There you will then see a series of emojis or numbers that you have to compare with each other.\n\nThe best way to do this is to meet up or start a video call. 👭",
     "verifyOtherDevice": "🔐 Verify other device",
--- a/lib/pages/settings_chat/settings_chat.dart
+++ b/lib/pages/settings_chat/settings_chat.dart
@@ -3,6 +3,7 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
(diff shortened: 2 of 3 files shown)


krille-chan:
Documentation smells like LLM. Please don't use LLMs anymore from now on to generate code or documentation. I don't think we to add a slider for this. We should rather decide for a good default and keep the setting in the advanced configuration

jonharris0n:
I understand the motivation, especially given the number of low-quality AI-generated PRs maintainers are seeing.

That said, I think this is a difficult line to draw and one that will only get harder over time. AI assistance is rapidly becoming part of the standard development toolchain, much like autocomplete, refactoring tools, and countless other productivity aids before it.

I’d rather see code judged on its quality, maintainability, and whether the contributor understands it than on how the first draft was produced. That feels like a more durable standard.

PR #3061, opened 2026-06-03
support stable m.poll.start event namespaces and add unit tests

mylacc:

- [X] Android
- [ ] iOS
- [X] Browser (Chromium based)
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [X] Desktop macOS

### DIFF
--- a/lib/pages/chat/events/poll.dart
+++ b/lib/pages/chat/events/poll.dart
@@ -3,9 +3,10 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import 'package:async/async.dart';
+import 'package:collection/collection.dart';
 import 'package:fluffychat/config/app_config.dart';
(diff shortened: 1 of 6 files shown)


lib/pages/chat/events/poll.dart
```diff
@@ -26,12 +27,99 @@ class PollWidget extends StatelessWidget {
     super.key,
   });
 
-  void _endPoll(BuildContext context) =>
-      showFutureLoadingDialog(context: context, future: event.endPoll);
+  Map<String, Set<String>> _getPollResponses(Timeline timeline) {
```
krille-chan, line 30: This is the wrong place to implement it. You should contribute to the Matrix Dart SDK instead

PR #3063, opened 2026-06-03
Add mobile app switcher privacy setting

### LINKED ISSUE #2453: Hide app content in app switcher/app overview screen
### Feature Description

Hide app content in the app switcher/app overview/recents screen in both android and IOS.

### Rationale

Despite app lock being enabled, IOS/android still captures and displays the app’s UI in the app switcher/app overview/recents screen which anyone can view. This bypasses and undermines the intended protection provided by the app lock.

Matrix clients Element X and SchildiChat Next already do this. I didn't test their legacy versions (element and SchildiChat). Whatsapp also does this. Fluffychat is the only odd one out.

### Mockup

_No response_

### Additional Context

https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE

https://developer.apple.com/documentation/uikit/preparing-your-ui-to-run-in-the-background#Release-resources-upon-entering-the-background

https://pub.dev/packages/secure_application

The other clients I mentioned do this indiscriminately. It might be a good idea to only do this if the app lock feature is enabled.

aaravrav:
Adds a mobile-only app switcher privacy mode with Off, App lock, and Always options, defaulting to Always. The setting hides app content from recents/app switcher views without blocking normal foreground screenshots.

On Android 13+, uses the platform recents screenshot API. On Android 12L and older, applies FLAG_SECURE only while the app is leaving the foreground and clears it again on resume. On iOS and other mobile lifecycle snapshots, shows a full-screen privacy cover while inactive, hidden, or paused.

Closes #2453 

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/android/app/src/main/kotlin/chat/fluffy/fluffychat/MainActivity.kt
+++ b/android/app/src/main/kotlin/chat/fluffy/fluffychat/MainActivity.kt
@@ -1,27 +1,64 @@
 package chat.fluffy.fluffychat
 
+import android.content.Context
+import android.os.Build
+import android.view.WindowManager
 import io.flutter.embedding.android.FlutterActivity
 import io.flutter.embedding.engine.FlutterEngine
-
-import android.content.Context
+import io.flutter.plugin.common.MethodChannel
 
 class MainActivity : FlutterActivity() {
+    private var appSwitcherPrivacyEnabled = false
+    private var appSwitcherPrivacyForeground = true
 
     override fun attachBaseContext(base: Context) {
         super.attachBaseContext(base)
     }
-
 
     override fun provideFlutterEngine(context: Context): FlutterEngine? {
         return provideEngine(this)
     }
 
     override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
-        // do nothing, because the engine was been configured in provideEngine
+        MethodChannel(
+            flutterEngine.dartExecutor.binaryMessenger,
+            APP_SWITCHER_PRIVACY_CHANNEL
+        ).setMethodCallHandler { call, result ->
+            when (call.method) {
+                "setState" -> {
+                    appSwitcherPrivacyEnabled = call.argument<Boolean>("enabled") ?: false
+                    appSwitcherPrivacyForeground = call.argument<Boolean>("foreground") ?: true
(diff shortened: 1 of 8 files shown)


android/app/src/main/kotlin/chat/fluffy/fluffychat/MainActivity.kt
```diff
@@ -1,27 +1,64 @@
...
         super.attachBaseContext(base)
     }
 
-
     override fun provideFlutterEngine(context: Context): FlutterEngine? {
         return provideEngine(this)
     }
 
     override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
-        // do nothing, because the engine was been configured in provideEngine
```
krille-chan, line 20: There are packages on pub.dev which can do this so I don't have to maintain it

aaravrav: Not really. The closest ones either keep FLAG_SECURE enabled while protection is active, which blocks foreground screenshots, or still require MainActivity lifecycle integration for app switcher only behavior.

The Android 13+ part of this PR uses setRecentsScreenshotEnabled(false), which none of the packages expose. That API is the reason screenshots remain allowed while Recents is hidden on Android 13+, and is the idiomatic way to do so.

I could if you prefer,
1. switch to a package even if that means foreground screenshots are blocked while enabled, or
2. first propose/add setRecentsScreenshotEnabled support upstream to a package such as screen_protector?

aaravrav: @krille-chan Let me know because if you prefer the 2nd option then it'll take time to upstream those changes in the flutter package as well

aaravrav: You there?

aaravrav:
This has been waiting without a response for a while. Can you please say which direction you want? I personally need this functionality for my phone to use fluffychat

Plus all other matrix clients already implement this

PR #3072, opened 2026-06-08
fix(chat): compact HTML message rendering

wdx-agent-io:
I noticed a few rough edges in `HtmlMessage` while testing markdown
rendering on Android:

1. Multi-line `<li>` lists collapsed to a single line — `<li>` only
   lived in `fullLineHtmlTag`, so consecutive items had no separator.
2. The block-element separator was `'\n\n'`, which gave two empty
   lines between paragraphs.
3. `<br>` was rendered as plain `'\n'`, so Flutter's line-height
   multiplier inflated every soft break.

The fix moves `<li>` to `allowedHtmlTags` (so it uses the
`blockHtmlTags` `'\n'` delimiter), reduces the block separator to
`'\n'`, and replaces `<br>` `'\n'` with a `WidgetSpan` of fixed
height (`fontSize * 0.4`).

Tested on Android 16 / Flutter 3.44, see attached screenshots.

PR #3073, opened 2026-06-08
fix: tighten HTML message rendering

wdx-agent-io:
Supersedes #3072 (closed by @krille-chan — Chinese code comments and
AI-style description).

I noticed a few rough edges in `HtmlMessage` while testing markdown
rendering on Android:

1. Multi-line `<li>` lists collapsed to a single line — `<li>` only
   lived in `fullLineHtmlTag`, so consecutive items had no separator.
2. The block-element separator was `'\n\n'`, which gave two empty
   lines between paragraphs.
3. `<br>` was rendered as plain `'\n'`, so Flutter's line-height
   multiplier inflated every soft break.

The fix moves `<li>` to `allowedHtmlTags` (so it uses the
`blockHtmlTags` `'\n'` delimiter), reduces the block separator to
`'\n'`, and replaces `<br>` `'\n'` with a `WidgetSpan` of fixed
height (`fontSize * 0.4`).

Tested on Android 16 / Flutter 3.44, see attached screenshots in
#3072 for reference.

### DIFF
--- a/lib/pages/chat/events/html_message.dart
+++ b/lib/pages/chat/events/html_message.dart
@@ -105,6 +105,7 @@ class HtmlMessage extends StatelessWidget {
     'table',
     'details',
     'blockquote',
+    'li',  // required so consecutive <li> items do not collapse onto one line
   };
 
   /// We add line breaks before these tags:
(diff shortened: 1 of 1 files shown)


krille-chan:
AI

Issue #1294, opened 2024-08-22
Always show time/ date of the messages


vyrgozunqk:
### Feature Description

Currently the only way to see when someone has sent specific message is to tap on it and select it.
It would be nice if there is at least toggle switch to always show inline when the message was sent and if it was today or some other day, it can be next to the chat bubble, or above it as Element does. 
Telegram has different approach and puts the time of the message inside the chat bubble.

### Rationale

This is useful for larger rooms, work environments, time critical tasks and etc., when going back through the chat logs.

### Mockup

<img width="459" alt="image" src="https://github.com/user-attachments/assets/97271089-1f99-45c7-ad42-5cfd412a1a5e">

youainti:
I would also like this feature because I use some chat rooms to track the timing of events.

ariqpradipa:
+1 for this! Im moving from WhatsApp and Telegram, which show the exact time for each message. It feels a bit strange not being able to see the precise time a message was sent, since here it’s grouped by time frames instead.

example on telegram:
<img width="489" height="180" alt="Image" src="https://github.com/user-attachments/assets/64b24358-cff1-45ac-9d90-8b7574f0d8d9" />

or, if any of you have ever use Line, their implementation is like this
<img width="489" alt="Image" src="https://github.com/user-attachments/assets/229654e7-2445-46e8-804d-c5f9a8cba8b9" />

wulfjoergensen:
+1 for this
need this for me and to get my friends switching to matrix

PR #3092, opened 2026-06-11
feat: Add per-message timestamps

PolynomialDivision:

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

This is the easiest way you can add timestamps to messages without breaking the coding style meaning using StatelessWidget and just override the "build". It even reduces code complexity from my point of view. I searched for a way to make it look like Element-X or WhatsApp, but this from my point of view is only possible with a custom RenderBox (I guess you would not accept that). Otherwise you have to reserve extra space in the bubble and you cant put the time to "half of the text" height. The issue I can reference here is https://github.com/krille-chan/fluffychat/issues/1294

### DIFF
--- a/lib/config/setting_keys.dart
+++ b/lib/config/setting_keys.dart
@@ -76,6 +76,7 @@ enum AppSettings<T> {
   chatFilter<String>('chat.fluffy.chat_filter', 'allChats'),
   hideRoomsInSpaces<bool>('chat.fluffy.hideRoomsInSpaces', false),
   showThumbnailsInTimeline<bool>('chat.fluffy.showThumbnailsInTimeline', true),
+  showMessageTimestamp<bool>('chat.fluffy.showMessageTimestamp', true),
   debugPush<bool>('chat.fluffy.debug_push', false);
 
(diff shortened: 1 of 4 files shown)


krille-chan:
Sorry, I disagree that we should solve this with another config option as we already have a lot of config creep. This does not fit into the design vision

PolynomialDivision:
So no timestamps per message at all, or doing a more complicated widget with custom renderbox to emulate Telegram/WhatsApp/Element-X? Or default true?

PolynomialDivision:
I did not compile and checked out the latest dev branch, but is that commit adding to each message a timestamp?
https://github.com/krille-chan/fluffychat/commit/a86fa3ccad0347686e556dfb872d556a9c4354ef

PolynomialDivision:
I installed the latest app, and indeed, it adds to each message a timestamp. Looks awesome! Thanks. I will close.

PR #3132, opened 2026-06-20
fix: fixes memory leaks due to missing .dispose() calls causing UI freeze

rippa-the-hutt:
Hello!
I was experiencing a few UI freezes while using your otherwise excellent chat program. They happen after a while that the sw has been kept running, and they seem to only be concerning the UI.
The main issue I found was that this snippet of file `lib/widgets/matrix.dart`:
`void dispose() {`
`      WidgetsBinding.instance.removeObserver(this);`
` `
`      onRoomKeyRequestSub.values.map((s) => s.cancel());   // ← NEVER EXECUTED`
`      onKeyVerificationRequestSub.values.map((s) => s.cancel()); // ← NEVER EXECUTED`
`      onLogoutSub.values.map((s) => s.cancel());            // ← NEVER EXECUTED`
`      onNotification.values.map((s) => s.cancel());          // ← NEVER EXECUTED`
`...`
Was not being executed as, according to [Dart's docs](https://api.dart.dev/dart-core/Iterable/map.html), Iterable.map() is lazy. Hence, being not consumed in this function, nothing was really being disposed of.
I used plain for loops that, while way less elegant than your solution, should get the job done.
There was also a minor leak in lib/pages/chat/chat.dart, but I think this might have been equally dangerous as it seems that every sync might leak memory, so I included a fix for that. I am currently running this patched version on my computer and it seems to be running fine (Arch Linux, had to modify an header include to build on GCC 16 but this PR is strictly concerned with fixing the memory leak. I don't think this modification impacts the issue).
I am sorry I can't test this on other systems but I am a bit of a dinosaur and don't have decent mobiles at my disposal, nor I have windows/apple systems.

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -591,7 +591,13 @@ class ChatController extends State<ChatPageWithRoom>
   void dispose() {
     timeline?.cancelSubscriptions();
     timeline = null;
+    _storeInputTimeoutTimer?.cancel();
+    typingCoolDown?.cancel();
+    typingTimeout?.cancel();
+    sendController.dispose();
+    scrollController.dispose();
     inputFocus.removeListener(_inputFocusListener);
+    inputFocus.dispose();
     web.window.removeEventListener('paste', _handleClipboardFilePasteWeb);
     if (currentlyTyping) room.setTyping(false);
     MxcImage.clearCache(widget.room.id);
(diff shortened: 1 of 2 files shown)


krille-chan (approved):
Thank you very much. Looks like I need to more carefully use everything that can be disposed and try to find a lint for it

rippa-the-hutt:
My pleasure!
I see that there are a few solutions to check this (like  running flutter with `--observe` + http://localhost:8100 , if you are allergic to IDEs as I am lol).
If I have some time tomorrow I will try this approach in my normal usage, see whether I can find some other leak. But this is now running really finely.
cya!
Rippa

PR #3153, opened 2026-06-23
feat: improve caption display for portrait images

PolynomialDivision:
If a portrait image has a caption, the text gets compressed to the width of the image which often looks bad. Expand to full bubble width and crop the image as a tradeoff. User needs to tap to see the full image.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/events/message_content.dart
+++ b/lib/pages/chat/events/message_content.dart
@@ -15,6 +15,7 @@ import 'package:matrix/matrix.dart';
 
 import '../../../config/app_config.dart';
 import '../../../utils/event_checkbox_extension.dart';
+import '../../../utils/file_description.dart';
 import '../../../utils/platform_infos.dart';
 import '../../../utils/url_launcher.dart';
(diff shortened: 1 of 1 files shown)


krille-chan:
> If a portrait image has a caption, the text gets compressed to the width of the image which often looks bad. Expand to full bubble width and crop the image as a tradeoff. User needs to tap to see the full image.

But for a picture with very small width and large height this would look super bad and pixelated, right?

PolynomialDivision:
> But for a picture with very small width and large height this would look super bad and pixelated, right?

I think it is not worse than before. A user needs to click on the image than... I think this is fine.

krille-chan:
@PolynomialDivision but by doing it fit cover, not everything of the image will be visible. How are other messengers solving this problem?

PR #3245, opened 2026-07-13
fix: ImageViewer and MxcImageViewer should not use root navigator

### LINKED ISSUE #2495: pressing back to close image, closes chat
### Bug Description

When trying to close an image by pressing back, the chat is closed but image stays open. 

### Steps to Reproduce

1. Click on an image to view it
2. Press back

### Expected Behavior

The image should be closed, and the chat stay open

### App Version

2.4.0

### Platform

Android (PlayStore)

fairphone 5 android 16

### Additional Context

_No response_

TehDust:
This resolves https://github.com/krille-chan/fluffychat/issues/2495
`showDialog` uses flutters root navigator which was conflicting with the navigator used by `go_router`. Setting `useRootNavigator` to false puts the dialog on the same navigator which lets the back gesture properly pop it.

I've applied and tested these changes on dialogs that use `ImageViewer` and `MxcImageViewer` as these are the ones the bug is concerned with. There are other instances where `showDialog` is used but I could not produce similar bugs. They don't use the image viewer.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/events/message_content.dart
+++ b/lib/pages/chat/events/message_content.dart
@@ -92,6 +92,7 @@ class MessageContent extends StatelessWidget {
               textColor: textColor,
               onTap: () => showDialog(
                 context: context,
+                useRootNavigator: false,
                 builder: (_) => ImageViewer(
                   event,
                   timeline: timeline,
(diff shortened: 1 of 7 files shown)


krille-chan:
grafik

This is not a fix because with this in column mode the image viewer is no longer filling the full window.

PR #3263, opened 2026-07-16
Add bidi support

mohsenasm:
Hi,

I added support for bidi.

Related issue: https://github.com/krille-chan/fluffychat/issues/63

## How it works

- I redefined `Text` by extending `material.Text`, fixing `textDirection` based on the input data.
- I also redefined `TextField` by wrapping `material.TextField`. It caches `textDirection`, refreshes it when the text changes, and passes it to the original widget.
- Instead of importing `package:flutter/material.dart` directly, we now import `package:fluffychat/widgets/bidi/material.dart`. This file exports our custom `Text` and `TextField`, along with everything else from the material package:

```dart
export 'package:fluffychat/widgets/bidi/text.dart';
export 'package:fluffychat/widgets/bidi/textfield.dart';
export 'package:flutter/material.dart' hide Text, TextField;
```

This keeps the rest of the code unchanged, aside from replacing material imports. We can also add a linter rule to prevent direct imports of `package:flutter/material.dart` in the future.

Screenshot before this change (wrong direction):
s1

Screenshot after this change (correct direction):
s2

Sorry for the large number of changed files — replacing the material import required it.

---

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


mohsenasm:
@krille-chan, if you have some time, I’d really appreciate it if you could take a look at this PR.

lib/widgets/bidi/text.dart
```diff
@@ -0,0 +1,48 @@
+import 'package:fluffychat/utils/bidi.dart';
+import 'package:flutter/material.dart' as material;
+
+class Text extends material.Text {
```
krille-chan, line 4: Inheritance instead of composition is considered a bad pattern in flutter. However I have the feeling that what you want to achieve can be achieved much simpler

lib/widgets/bidi/text.dart
```diff
@@ -0,0 +1,48 @@
...
+    super.textScaler,
+    super.maxLines,
+    super.semanticsLabel,
+    super.textWidthBasis,
+    super.textHeightBehavior,
+    super.selectionColor,
+  }) : super.rich();
+
+  @override
+  material.TextDirection? get textDirection {
```
krille-chan, line 40: so if your only goal is to overwrite textDirection you can just use this: https://stackoverflow.com/questions/60846669/how-to-set-textdirection-to-a-flutter-theme instead of overwriting **all** Text widgets which is super overkill. However why do we need to do this? LTR text is automatically detected and already enabled:

Image

mohsenasm: I don’t think we should set text direction for the entire app, because that would make direction depend on the system locale. Many people use English as their phone’s main language but still write in RTL languages.

This should be handled per `Text` instead. One message can be LTR (for example, English), while another can be RTL (for example, Persian). The same applies to any user-entered text, such as room names.

Please see https://github.com/krille-chan/fluffychat/issues/63 and https://github.com/krille-chan/fluffychat/pull/1239. Telegram handles it this way as well.

Right now, typing and reading RTL text is difficult. When an English word is mixed into RTL content, the sentence gets scrambled.

lib/widgets/bidi/text.dart
mohsenasm: Should we create a `Text` class that wraps `material.Text`, similar to what we did with [TextField](https://github.com/krille-chan/fluffychat/pull/3263/changes/BASE..84a69d2bd692559c9f439ef2e16b45e7bc2d6be8#diff-dfac5d65f74b7874e9e7fb0b3ebbf5504ce756dc59890983bed50f5635fbeca2)?

lib/widgets/bidi/text.dart
krille-chan: > I don’t think we should set text direction for the entire app, because that would make direction depend on the system locale. 

No, you can set it by yourself and make this configurable

mohsenasm: Hi again,
Please see the screenshots below:

| Current Version (English) | Current Version (Arabic) |
| --- | --- |
| fdroid-en1 |  fdroid-ar1 |

| This PR (English) | This PR (Arabic) |
| --- | --- |
|  my-pr-en1 |  my-pr-ar1 |

As shown above, bidi handling should work correctly regardless of the app language. Users of RTL languages often exchange mixed-direction text, and these messages may start with either English words or RTL words. The UI should render both cases properly.

At the moment, it looks like the app only handles one direction correctly at a time. When the app is set to an RTL language, messages that start with English text are displayed incorrectly. Likewise, when the app is set to an LTR language, messages that start with RTL text are rendered incorrectly.

The issue is that any message, in any room, may contain either direction of text, and the rendering should handle that correctly.

---

There was a previous PR from two years ago that proposed handling this by manually calling `.bidiFormatted` for text widgets in the codebase: https://github.com/krille-chan/fluffychat/pull/1239

That approach was not accepted because it would require developers to remember to apply `.bidiFormatted` each relevant place, which is easy to miss and not scalable.

This PR automatically detects and corrects bidi text for all texts, so mixed LTR/RTL content is handled consistently without relying on manual changes throughout the codebase.

krille-chan:
This approach makes no sense to me. There is no need to overwrite `Text` widget. RTL/LTR can be done with the `Directionality` widget globally: https://leancode.co/blog/right-to-left-in-flutter-app

PR #3314, opened 2026-07-24
feat: add double-tap to react option in settings

AndusDEV:
Adds an optional double-tap message reaction feature. This is a familiar gesture from popular messengers like Messenger (which I'm migrating from currently), WhatsApp and Telegram.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/config/setting_keys.dart
+++ b/lib/config/setting_keys.dart
@@ -75,7 +75,9 @@ enum AppSettings<T> {
   webNotificationSound<bool>('chat.fluffy.web_notification_sound', true),
   chatFilter<String>('chat.fluffy.chat_filter', 'allChats'),
   hideRoomsInSpaces<bool>('chat.fluffy.hideRoomsInSpaces', false),
-  showThumbnailsInTimeline<bool>('chat.fluffy.showThumbnailsInTimeline', true);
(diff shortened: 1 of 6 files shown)


krille-chan (commented):
Hm, I need to think about it. Double tap does not work in all messengers. For example in Google Messages there is nothing like that. Also right now on my phone I can double tap to select text which would get lost then. Yes we could make it configurable but then we would have another config creep. Probably it is not hard to maintain at least

AndusDEV:
> Hm, I need to think about it. Double tap does not work in all messengers. For example in Google Messages there is nothing like that. Also right now on my phone I can double tap to select text which would get lost then. Yes we could make it configurable but then we would have another config creep. Probably it is not hard to maintain at least

No worries at all! I actually accounted for text selection when building this. If the setting is toggled off, double-tap still works as standard text selection.

AndusDEV:
Just a quick update: I realized I forgot to sign my commit, so I've amended and force-pushed it. Let me know if you have any more thoughts on the text selection behavior or if there's anything else you'd like me to tweak.

krille-chan (approved):

PR #3317, opened 2026-07-25
fix: notification reactions not shown in unifiedpush, fcm worked

PaulaBras:
Notification actions (reply, mark as read, mute) were suppressed on the UnifiedPush path as a workaround for [flutter-connector#34](https://codeberg.org/UnifiedPush/flutter-connector/issues/34); that bug was fixed upstream in unifiedpush_android 3.4.1, which pubspec.lock already pins, so the workaround is no longer needed.

Background Info:
This resulted in Android Auto not working with Fluffy since (to my understanding) Android Auto only catches chat messages, which they were not classified as without reactions.

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/utils/background_push.dart
+++ b/lib/utils/background_push.dart
@@ -433,8 +433,6 @@ class BackgroundPush {
       l10n: l10n,
       activeRoomId: matrix?.activeRoomId,
       flutterLocalNotificationsPlugin: _flutterLocalNotificationsPlugin,
-      useNotificationActions:
-          false, // Buggy with UP: https://codeberg.org/UnifiedPush/flutter-connector/issues/34
     );
   }
 }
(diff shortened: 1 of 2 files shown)


krille-chan:
Have you actually tested this?

PaulaBras:
> Have you actually tested this?

yes. Not the Android Auto issue, but the push notification arrived with reaction this time. I tested it with sunup (mozilla)

krille-chan:
@PaulaBras Your commit is not signed. This is point 6 of https://github.com/krille-chan/fluffychat/blob/main/CONTRIBUTING.md#contributing-guidelines

PaulaBras:
@krille-chan the commit is signed using my ssh key. name of the commit was "wrong" according to guidelines. this is fixed now.

krille-chan (approved):

krille-chan:
@PaulaBras thank you so much :-)

PR #3323, opened 2026-07-27
feat(android): support sharing images, video, audio, and files

mikhail-angelov:
- Updated AndroidManifest to replace generic text/* and */* MIME filters with specific types (image, video, audio, application) for both single and multiple share intents.
- Added onNewIntent override in MainActivity to handle incoming image shares: copies shared image to app cache before URI permission is lost, then invokes a MethodChannel (`fluffychat/share`) with the file path.
- Removed general */* intent filter to avoid unintended handling of non-media types while keeping support for text/plain.

- [*] I have read and understood the [contributing guidelines](https://github.com/krille-chan/fluffychat/blob/main/CONTRIBUTING.md). 

- [*] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -80,23 +80,25 @@
                <data android:scheme="im.fluffychat" android:host="chat" />
             </intent-filter>
 
-            <!-- App can receive shared text -->
+            <!-- App can receive shared content (text, images, video, audio, files) -->
             <intent-filter>
                <action android:name="android.intent.action.SEND" />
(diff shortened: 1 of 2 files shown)


krille-chan:
This is already implemented by hsing the receive_sharing_intent package

mikhail-angelov:
oh, sorry, I missed it, I'll update code with new PR

PR #3324, opened 2026-07-27
refactor(android): restrict share intent filters to specific mime types

mikhail-angelov:
Replace generic `text/*` and wildcard `*/*` mime type filters with explicit listings for text/plain, image/*, video/*, audio/*, and application/* for both single `SEND` and multiple `SEND_MULTIPLE` actions. This ensures the app only receives supported content types, improving security and predictability.

- [*] I have read and understood the [contributing guidelines](https://github.com/krille-chan/fluffychat/blob/main/CONTRIBUTING.md). 

- [*] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


krille-chan:
Why is this a problem? Which kind of file is not supported? If it is an unknown file type then it would just send it as a file? There is no issue for this

mikhail-angelov:
> Why is this a problem? Which kind of file is not supported? If it is an unknown file type then it would just send it as a file? There is no issue for this

I does not support **Share** action on Camera app or Image Gallery. once I want share camera photo I do not see Fluffy as application to share it., the same thing with video files.

krille-chan:
> > Why is this a problem? Which kind of file is not supported? If it is an unknown file type then it would just send it as a file? There is no issue for this
> 
> I does not support **Share** action on Camera app or Image Gallery. once I want share camera photo I do not see Fluffy as application to share it., the same thing with video files.

I cannot reproduce it. I've tried on three different android devices, with stock android, Nothing OS and One UI and on all devices FluffyChat appears as an application to share it from Camera app and Image Gallery. So maybe it's a bug with your phone?

mikhail-angelov:
> > > Why is this a problem? Which kind of file is not supported? If it is an unknown file type then it would just send it as a file? There is no issue for this
> > 
> > 
> > I does not support **Share** action on Camera app or Image Gallery. once I want share camera photo I do not see Fluffy as application to share it., the same thing with video files.
> 
> I cannot reproduce it. I've tried on three different android devices, with stock android, Nothing OS and One UI and on all devices FluffyChat appears as an application to share it from Camera app and Image Gallery. So maybe it's a bug with your phone?

maybe you are right, I have pretty old Huawei on Android 10, so if it's not reproducible on new phones, let's close the ticket, and I'll just keep my local change.

Issue #3181, opened 2026-06-28
The emoji/sticker selection panel is too high.


Quffik:
### Bug Description

On my two devices with different aspect ratios, the emoji/sticker selection bar is too tall. Sorry if someone has already reported this, I couldn't find the report.

This is my first Issue on GitHub.

### Steps to Reproduce

1. Go to chat.
2. Open the emoji/sticker selection panel.

<!-- Uploading "IMG_20260628_221453_791.jpg"... --

### Expected Behavior

I go into the chat and try to send a sticker/emoji.

### App Version

2.7.2/2.6.0

### Platform

Android (Other)

### Additional Platform Information

Pixel 7 Pro

Quffik:
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/cd8ffee7-b64c-43b6-b574-f12029425396" />
<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/8557d4ae-6ff1-48c1-97c2-e0c27f254fc5" />

PR #3347, opened 2026-08-02
fix: remove extra spacing in sticker selection dialog

ludmila-lovelace:
Fixes an issue I started to see recently where there's some extra spacing in the sticker selection dialog, which was also reported in #3181. See below for comparison on Android/Desktop Linux:

|Before| After|
|---|---|
| Screenshot_1785633195 | Screenshot_1785633206 |
| image | image |

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/sticker_picker_dialog.dart
+++ b/lib/pages/chat/sticker_picker_dialog.dart
@@ -39,10 +39,12 @@ class StickerPickerDialogState extends State<StickerPickerDialog> {
     return Material(
       color: theme.colorScheme.onInverseSurface,
       child: SafeArea(
+        top: false,
         child: CustomScrollView(
           slivers: <Widget>[
             SliverAppBar(
               floating: true,
+              primary: false,
               toolbarHeight: 72,
               scrolledUnderElevation: 0,
               backgroundColor: Colors.transparent,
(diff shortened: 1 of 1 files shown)


lib/pages/chat/sticker_picker_dialog.dart
```diff
@@ -180,3 +180,32 @@ class StickerPickerDialogState extends State<StickerPickerDialog> {
     );
   }
 }
+
+class SearchBarSliverDelegate extends SliverPersistentHeaderDelegate {
```
krille-chan: Using inhertience instead of composition is considered a bad pattern in Flutter: https://stackoverflow.com/questions/74540642/when-should-you-prefer-widget-inheritance-over-composition-in-flutter can you use composition instead?

ludmila-lovelace: @krille-chan thanks for taking a look so quickly! I had a look at the original `SliverAppBar` implementation and I think inheritance is unavoidable since `SliverPersistentHeaderDelegate` is an abstract class? In doing so though, I found a much easier fix that avoids needing to use `SliverPersistentHeader` altogether so will push that instead.

krille-chan: That looks great! Thank you so much ♥️

krille-chan (approved):

PR #3427, opened 2026-08-20
Trust user installed CA certificates on Android

### LINKED ISSUE #1393: Support self signed certificates
### Feature Description

I would like to see Fluffychat to support self signed certificates. I have my own Certificate Authority on a server, and just deploy main CA.crt to all devices on my LAN. I can connect to matrix https from Element Android and Element UI, but fluffy chat responds with No connection to server if you use **https://dietpi.lan:5335** (synapse host:port). When I disable HTTPS and switch to http **http://dietpi.lan:5335**, its working OK. I am using nginx reverse proxy

Screenshot_20241007-224032

Screenshot_20241007-224443

### Rationale

Enables TLS on LAN thus securing Matrix connection between server and clients

### Mockup

_No response_

### Additional Context

_No response_

Belkaar:

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

This pull request adds the user certificates to the trusted CAs to the custom http client.
Fixes https://github.com/krille-chan/fluffychat/issues/1393

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -42,6 +42,7 @@ dependencies:
   flutter_new_badger: ^2.0.0
   flutter_secure_storage: ^11.0.0
   flutter_shortcuts_new: ^2.0.0
+  flutter_user_certificates_android: ^0.1.0
   flutter_vodozemac: ^0.8.1
   flutter_web_auth_2: ^5.1.0
   flutter_webrtc: ^1.6.0
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -42,7 +42,6 @@ dependencies:
   flutter_new_badger: ^2.0.0
   flutter_secure_storage: ^11.0.0
   flutter_shortcuts_new: ^2.0.0
-  flutter_user_certificates_android: ^0.1.0
   flutter_vodozemac: ^0.8.1
   flutter_web_auth_2: ^5.1.0
   flutter_webrtc: ^1.6.0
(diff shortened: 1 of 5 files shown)


krille-chan:
Not sure if we really need this in a world where we have LetsEncrypt. Can't you just make sure you have a valid trust chain?

Belkaar:
This is usefull for two scenarios:
1) LAN-Solution where you dont have or want internet access to constantly renew certificates so you use a self signed CA.
2) Old devices where letsencrypt no longer works. There you can just install the current Letsencrypt root ca on the device without rooting.

Anyways: It does not cost anything to have this and adds flexibility. User CAs can only be installed by the user via settings and not by apps, so I dont see a security concern. Most devices even show a small reminder in the notifications pull down.

pubspec.yaml
```diff
@@ -41,6 +41,7 @@ dependencies:
   flutter_new_badger: ^2.0.0
   flutter_secure_storage: ^11.0.0
   flutter_shortcuts_new: ^2.0.0
+  flutter_user_certificates_android: ^0.1.0
```
krille-chan: I don't want to add another dependency for this

Belkaar:
I will update this to not need an additional dependency

pubspec.yaml
Belkaar: i removed the dependency and added code to load the certs manually

PR #3430, opened 2026-08-21
fix: remove extra spacing in autocomplete popover

ludmila-lovelace:
- Similar to previous fix, this one is for removing the extra spacing at the top of the autocomplete popover that displays for room commands and custom emotes
- The issue was visible on Android and iOS but not Linux
- Fix works for Android and has no effect on existing Linux UI
- Comparison below (popover is red for clarity)

|Before| After|
|---|---|
| Screenshot_1787283665 | Screenshot_1787283771 |

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


krille-chan (approved):
TYSM

Issue #290, opened 2023-07-15
Feature Request: Make emoji-reactions searchable with text typing.


krille-chan:
#### Description

FluffyChat supports many emoji symbols for reactions. But it is often tedious to find the specific symbol.

Here I wish that fluffychat offers a text input field where for that symbol can be searcheg (e.g. `:raised_hands:`).

#### Additional information:
- Device: *GPD MicroPC (version 1)*
- OS and OS version: *Artix GNU/Linux (rolling release)*
- Installed version of FluffyChat: *1.2.0.r14.gffaab465* (git checkout)

compuguy:
That's disappointing to hear. Other matrix clients have *some sort* of emoji search function....

tmayoff:
I'd also like to see this implemented. It looks like they started on this but the PRs were closed due to being stale. https://github.com/krille-chan/fluffychat/pull/454 (and maybe https://github.com/krille-chan/fluffychat/pull/455)

Soundtoxin:
This is a weak point of many emoji pickers (particularly with reactions where you can't even paste in an emoji from elsewhere as a workaround like you can with a message), Element X and Conversations (XMPP) have the same problem, but the old Element for Android is very nicely done with a search bar in the picker that gets focused automatically and opens the keyboard. May be worth a look.

PR #3450, opened 2026-08-23
feat: use custom emoji picker widget and add search

### LINKED ISSUE #3339: The emoji picker should be searchable
### Feature Description

Hi all,

it would be nice if emojis could be queried by search terms in the emoji picker, like e.g. in Signal.

Actually, Fluffycat even already contains the needed data, at least partially, the [sas-emoji.json](https://github.com/matrix-org/matrix-spec/blob/main/data-definitions/sas-emoji.json) definition file that describes or "translates" the emojis, but it's IIRC only used for verification. Maybe one would have to add the (more complete?) [cldr annotations](https://github.com/unicode-org/cldr/tree/main/common/annotations) or use them instead.

### Rationale

Other messengers, esp. Signal have this, one can search for emojis. Would be nice to have, esp. for rarely used emojis not listed in the recently used list.

### Mockup

_No response_

### Additional Context

Thanks for maybe considering adding this :-)

ludmila-lovelace:
- Closes #3339 and maybe #290 (? not sure if this one is referring to the existing `:emote:` functionality or searchable emoji reacts)
- Move emoji picker dialog logic to separate class
- Use `customWidget` property to define own widget for emoji picker
- Replace recent emoji tab with custom widget
  - allows searching all emojis and displaying results
  - shows recents when first opened
  - clears search field when switching tabs
- Also add `cacheKey` to image reactions under messages to fix issue where reactions weren't updating correctly after being removed by user

| No Recents Page | Search Page |
|---|---|
| Screenshot_1787457772 | Screenshot_1787457785 |

| Recents Page | Search Page (as a reaction) |
|---|---|
| Screenshot_1787457807 | Screenshot_1787457820 |

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat_emoji_picker.dart
+++ b/lib/pages/chat/chat_emoji_picker.dart
@@ -3,9 +3,9 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import 'package:emoji_picker_flutter/emoji_picker_flutter.dart';
 import 'package:fluffychat/config/themes.dart';
 import 'package:fluffychat/l10n/l10n.dart';
+import 'package:fluffychat/pages/chat/emoji_picker_dialog.dart';
 import 'package:fluffychat/pages/chat/sticker_picker_dialog.dart';
 import 'package:fluffychat/pages/chat/trust_user_key_dialog.dart';
 import 'package:material_ui/material_ui.dart';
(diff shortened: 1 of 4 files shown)


ludmila-lovelace:
haiii @krille-chan, i think this one's at a point that i'm comfy with. just let me know if you want anything changed, i tried to make sure the controller/view widgets were done properly etc.

PR #3513, opened 2026-08-28
Fix a bunch of cases where the read marker wasn't being updated

ruka-hamanasu:
These cases would cause the dark blue dot to remain on a room (non-notifying events):

- Push-rule-gated rooms: Read markers are no longer blocked when a room’s push rules match nothing (e.g. channels set to mentions-only). The marker targets the latest real message-like event instead.
- Stuck unread indicator: If room.hasNewMessages is still true despite room.fullyRead already pointing at the latest event, the marker is re-posted to refresh receipt timestamps and clear the channel unread state. Note that this only happens if you are using another client such as Cinny that posts `m.read` or `m.private_read` but not `m.fully_read`.
- Failed marker requests no longer wedge the room: _setReadMarkerFuture is released in whenComplete() even when the request fails.
- Detached views not treated as at-bottom: When no scroll client is attached, the function falls back to the cached _scrolledUp flag instead of assuming the user is at the latest message.
- Context slices protected: When viewing a sliced timeline (timeline.allowNewEvent == false), the newest loaded event is not treated as the room’s latest and the marker is not auto-advanced.

Tested by doing all the stuff CI does locally on my machine and by running an APK build on my phone. My homeserver/account are highly prone to this issue and I had several stuck rooms previously.

This interacts with a couple of Tuwunel bugs that I have also put up PRs for: https://github.com/matrix-construct/tuwunel/pull/564 and https://github.com/matrix-construct/tuwunel/pull/565 . Also now https://github.com/matrix-construct/tuwunel/pull/566 which only impacts rooms with only backfilled events (like when the user is the first joiner from the local homeserver).

I am happy to split out the Cinny interaction fix into a separate PR if you like; that's the most impactful one for people who use multiple clients.

- [X] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [ ] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -596,11 +596,12 @@ class ChatController extends State<ChatPageWithRoom>
     // We are already setting a read marker
     if (_setReadMarkerFuture != null) return;
 
-    // We only set read marker if we are at the bottom
-    if (_scrolledUp) return;
-
-    // We do not set read marker if we offer user the scroll up banner
-    if (scrollUpBannerEventId != null) return;
+    // We only set read marker if we are at the bottom. Check the live scroll
+    // position, not the cached _scrolledUp flag, which can be stuck when no
+    // scroll transition ever fired.
+    if (scrollController.hasClients && scrollController.position.pixels > 0) {
+      return;
+    }
 
     // We do not set read marker if timeline is empty
     final timeline = this.timeline;
@@ -614,16 +615,18 @@ class ChatController extends State<ChatPageWithRoom>
     }
 
     final setOnLatestEvent = eventId == null;
+    // Target the latest message-like event. Do not gate this on push rule
+    // evaluation: read markers track what the user has seen, not what
+    // notifies them.
     eventId ??= timeline.events
         .firstWhereOrNull(
-          (event) => room.pushRuleState == PushRuleState.notify
-              ? room.client.pushruleEvaluator.match(event).notify
-              : {
-                      EventTypes.Message,
-                      EventTypes.Encrypted,
-                      EventTypes.Sticker,
-                    }.contains(event.type) &&
-                    event.eventId.isValidMatrixIdStrict(),
+          (event) =>
+              {
+                EventTypes.Message,
+                EventTypes.Encrypted,
+                EventTypes.Sticker,
+              }.contains(event.type) &&
+              event.eventId.isValidMatrixIdStrict(),
         )
         ?.eventId;
 
@@ -638,11 +641,12 @@ class ChatController extends State<ChatPageWithRoom>
     // unread indicator depends on.
     if (room.fullyRead == eventId && !room.hasNewMessages) return;
 
-    // Set a readmarker on a specific event, not latest, but room is not unread
-    // at all.
+    // Room shows nothing new, but the read marker still lags behind this
+    // event (e.g. receipts were posted by another device): advance it.
     if (setOnLatestEvent &&
         !room.hasNewMessages &&
-        room.notificationCount == 0) {
+        room.notificationCount == 0 &&
+        room.fullyRead == eventId) {
       return;
     }
 
@@ -653,7 +657,11 @@ class ChatController extends State<ChatPageWithRoom>
           eventId: eventId,
           public: AppSettings.sendPublicReadReceipts.value,
         )
-        .then((_) {
+        .catchError((Object e, StackTrace s) {
+          Logs().w('Failed to set read marker', e, s);
+        })
+        .whenComplete(() {
+          // Always release the latch, even when the request fails.
           _setReadMarkerFuture = null;
         });
   }
--- a/lib/pages/chat/chat.dart
+++ b/lib/pages/chat/chat.dart
@@ -633,8 +633,10 @@ class ChatController extends State<ChatPageWithRoom>
     // This is a sending event, we do not set a readmarker yet
     if (eventId.isValidMatrixIdStrict() == false) return;
 
-    // Already set a read marker on this event
-    if (room.fullyRead == eventId) return;
+    // Already set a read marker on this event, unless the room still shows
+    // new messages: re-posting refreshes the receipts' timestamps, which the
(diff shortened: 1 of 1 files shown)


lib/pages/chat/chat.dart
```diff
@@ -596,11 +596,12 @@ class ChatController extends State<ChatPageWithRoom>
...
-    if (_scrolledUp) return;
-
-    // We do not set read marker if we offer user the scroll up banner
-    if (scrollUpBannerEventId != null) return;
+    // We only set read marker if we are at the bottom. Check the live scroll
+    // position, not the cached _scrolledUp flag, which can be stuck when no
+    // scroll transition ever fired.
+    if (scrollController.hasClients && scrollController.position.pixels > 0) {
+      return;
+    }
```
ruka-hamanasu, line 606: This one might be too aggressive.

lib/pages/chat/chat.dart
```diff
@@ -614,16 +615,18 @@ class ChatController extends State<ChatPageWithRoom>
...
-                      EventTypes.Sticker,
-                    }.contains(event.type) &&
-                    event.eventId.isValidMatrixIdStrict(),
+          (event) =>
+              {
+                EventTypes.Message,
+                EventTypes.Encrypted,
+                EventTypes.Sticker,
+              }.contains(event.type) &&
+              event.eventId.isValidMatrixIdStrict(),
```
ruka-hamanasu, line 637: This one is almost certainly correct and is why the dark blue dot would stay while the light blue dot would not.

lib/pages/chat/chat.dart
```diff
@@ -633,14 +636,17 @@ class ChatController extends State<ChatPageWithRoom>
     // This is a sending event, we do not set a readmarker yet
     if (eventId.isValidMatrixIdStrict() == false) return;
 
-    // Already set a read marker on this event
-    if (room.fullyRead == eventId) return;
+    // Already set a read marker on this event, unless the room still shows
+    // new messages: re-posting refreshes the receipts' timestamps, which the
+    // unread indicator depends on.
+    if (room.fullyRead == eventId && !room.hasNewMessages) return;
```
ruka-hamanasu, line 650: This may not be absolutely necessary, but if fullyRead is up to date and the room still shows it has messages, there is no harm in re-posting.

lib/pages/chat/chat.dart
```diff
@@ -651,7 +657,11 @@ class ChatController extends State<ChatPageWithRoom>
           eventId: eventId,
           public: AppSettings.sendPublicReadReceipts.value,
         )
-        .then((_) {
+        .catchError((Object e, StackTrace s) {
+          Logs().w('Failed to set read marker', e, s);
+        })
+        .whenComplete(() {
+          // Always release the latch, even when the request fails.
```
ruka-hamanasu, line 671: This would cause FluffyChat to stop updating the read marker if there was an error.

lib/pages/chat/chat.dart
```diff
@@ -633,14 +636,17 @@ class ChatController extends State<ChatPageWithRoom>
...
 
-    // Set a readmarker on a specific event, not latest, but room is not unread
-    // at all.
+    // Room shows nothing new, but the read marker still lags behind this
+    // event (e.g. receipts were posted by another device): advance it.
     if (setOnLatestEvent &&
         !room.hasNewMessages &&
-        room.notificationCount == 0) {
+        room.notificationCount == 0 &&
+        room.fullyRead == eventId) {
```
ruka-hamanasu: If the room is not unread, I don't think there's a reason not to advance the marker.

ruka-hamanasu:
I'm going to try to figure out the minimal set of changes that fixes the problems, since I think a couple of these are too aggressive and could mark rooms as read when there are genuinely unread messages in them.

ruka-hamanasu:
Ok, I've made it less aggressive about updating the read marker, so it shouldn't mark events are read that the user hasn't actually seen or at least scrolled past.

krille-chan:
@ruka-hamanasu can you please tell me how to reproduce the original problem on main branch? Because every time I mark a room as read on main branch it works perfectly fine so I thought the problem was already fixed

ruka-hamanasu:
> @ruka-hamanasu can you please tell me how to reproduce the original problem on main branch? Because every time I mark a room as read on main branch it works perfectly fine so I thought the problem was already fixed

the repro is: on any Continuwuity or Tuwunel homeserver, use FluffyChat in rooms that are set up as "mentions only", wait for dark blue dot to appear, read room, observe dark blue dot never goes away.

DMed you a reg token for my homeserver which has a couple additional fixes for cases where Tuwunel's behavior exacerbated the problems, but they cover 1 niche case (room with only backfilled events), a case where if someone reacts to something while you're reading the room won't be marked read, and an interaction with other clients where the read marker can move backwards. This fix should show improvement on any Tuwunel or Continuwuity server because it makes FluffyChat advance the read marker even if there are only non-notifying events (messages but no mentions) in the room. Not sure why same doesn't happen with Synapse honestly (or even that it doesn't since I don't have a Synapse account).

ruka-hamanasu:
Due to a report by a user who *only* uses FluffyChat that this wasn't happening to them, I investigated possible interactions with other clients. I've confirmed that the problem only occurs if you are *also* using a client that doesn't post `m.fully_read`. Cinny only posts `m.read` or `m.private_read`, never `m.fully_read`. So Cinny interacts fine with Cinny, and FluffyChat doesn't mess up Cinny, but Cinny messes up FluffyChat because FluffyChat ends up in a situation where `hasNewMessages` is true but `m.fully_read` already points at the last event in the room. In such a case, the only way to clear the dark blue dot is to post a fresh own receipt to update the timestamp.

On the bright side, that means you don't need a Tuwunel server to reproduce this; that was a red herring. All you need to do is to log Cinny into your account, read a few unread rooms with Cinny, then go and try to read those same rooms from FluffyChat.

Issue #3272, opened 2026-07-19
Predictive back gestures causes windows like chat and settings close even when a popup is present


iWisp360:
### Bug Description

I've found a pretty annoying issue when using fluffychat on android 16, which has predictive back gestures by default. 

When you do the gesture of sliding to the side from the border of the screen, the opened window(the chat widget, the settings widget, or any widget that isn't the home widget) is closed instead of a popup if existing, like the image viewer or a context menu.

The offended parts of the app are every window but the home window, the main screen, where the chats are displayed, including:

- Chat window
- Settings window
- add account window
- stickers window

I think these widgets have that issue:

- `PopupMenuButton`
- Probably `ImageViewerView`

Noticed normal dialogs you'd show with showDialog() don't have this issue. however, showAboutDialog() has the issue.

I wrote flutter code, so I know my way around it. Maybe I could help with some code.

I'll send a video preview as soon as possible.

### Steps to Reproduce

1. If you are on android 15, go to developer options and enable predictive back gestures. on android 16 it should be enabled by default.
2. Enable gestures navigation.
3. Open fluffychat on any chat.
4. Get an image and open it, or go to the chat settings, and open the context menu. Both of these should open a popup in fluffychat.
5. Try to close the popup by doing the gesture of going back.
6. Instead of closing the popup, you close the window behind the popup, also rendering the still active popup's actions useless.

### Expected Behavior

when you open an image, or any context menu, the back gesture with predictive back gestures should close the active image or menu instead of the window behind the popup.

### App Version

2.7.2

### Platform

Android (F-Droid)

### Additional Platform Information

Redmi Note 12. OS: crDroid 12.7, based on AOSP

krille-chan:
Isnt this already reported somewhere?

iWisp360:
> Isnt this already reported somewhere?

Here #2709, but it seemed the issue was forgotten, since u marked it as non-reproducible. But the bug is still there. Anyways, gotta see why is this happening, though normally I wouldn't get this on a plain flutter app with PopupMenuButton, so it isn't a flutter bug.

PR #3531, opened 2026-08-30
fix pages closing with dialogs opened instead of dialogs closing. Fixes #2709, #3272, #2495

### LINKED ISSUE #2709: Back gesture with keyboard open closes chat instead
### Bug Description

Not sure what's going on here. But when the keyboard is open, and I swipe a "back" system gesture, it closes the chat rather than the keyboard.

It's pretty inconsistent though. Best repro step I could manage. Step 6 in particular is unsure. Haven't checked on other phones and with non-gesture system nav.

### Steps to Reproduce

1. Open a chat
2. Bring keyboard up by tapping input field
3. Type
4. Erase or send
5. Type again but don't send? Don't need to erase?
6. Use "back" system gesture to close keyboard
8. Chat is closed instead

Might need to repeat that cycle of typing and closing kb a few times

### Expected Behavior

Keyboard is closed, chat remains open, in all circumstances where the gesture is used when keyboard is open.

### App Version

2.4.0

### Platform

Android (PlayStore)

Samsung A53, Android 16, One UI 8

### Additional Context

_No response_

iWisp360:
Fixes #2709, #3272, #2495

- [x] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

The reason for this issue is that modals are being called imperatively, so go_router gets confused and predictive back gestures closes the window instead of the modal.

Disabling predictive back gestures on the app completely is a workaround for this issue.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -34,6 +34,7 @@
         android:allowBackup="false"
         android:fullBackupContent="false"
         android:localeConfig="@xml/locale_config"
+        android:enableOnBackInvokedCallback="false"
     >
         <activity
             android:name=".MainActivity"


iWisp360:
While dialogs have been fixed, I still have to implement this on other modals which has the same problem.

krille-chan:
I don't think it is a sufficient way to fix a "minor ux bug" by implementing a complete new state management framework which changes the whole app 😢 also please read the contribution guidelines!

iWisp360:
Removed riverpod completely and opted to native flutter solutions instead

iWisp360:
> I don't think it is a sufficient way to fix a "minor ux bug" by implementing a complete new state management framework which changes the whole app 😢 also please read the contribution guidelines!

the simplest approach is disabling predictive back gestures, literally a one liner on the Android Manifest

krille-chan:
> the simplest approach is disabling predictive back gestures, literally a one liner on the Android Manifest

How are you sure this does not have any unexpected side effects?

iWisp360:
> > the simplest approach is disabling predictive back gestures, literally a one liner on the Android Manifest
> 
> How are you sure this does not have any unexpected side effects?

Such as? Unless you explicitly depend on this feature on the app, the answer is no, ofc.

Predictive back gestures is a feature from Android 13 or 14 and upwards that allows you to see the previous screen while doing the back gesture, giving you a peek. 

Disabling it makes the back gesture act like any normal app on any platform, you just... Go back, no peek, no nothing.

iWisp360:
Btw, A fix for this other than disabling the android feature can be made, but it would require a way of wrapping the widgets in a pop scope so they don't go back, or a shell route wrapping the entire app in a stack so modals can be placed on top and inhibit gestures from the app when modals are present, idk.

I tried the first approach, but it would easily change more than 3000 lines on the prjkt just for that, cuz for that you had to change all pages and modals.

The stack approach may work, but it most likely leads to custom showDialog, showModalBottomSheet, etc. functions.

krille-chan:
> Predictive back gestures is a feature from Android 13 or 14 and upwards that allows you to see the previous screen while doing the back gesture, giving you a peek.

Yes and I don't want to loose this

iWisp360:
> > Predictive back gestures is a feature from Android 13 or 14 and upwards that allows you to see the previous screen while doing the back gesture, giving you a peek.
> 
> Yes and I don't want to loose this

ooff, then I gotta work to actually fix it. I'm going to draft this PR as I'm focusing on other stuff on my flutter app.

krille-chan:
@iWisp360 I really doubt that a fix would be over 3000 LoC changes. Please don't do that hassle. Sounds for me like you are on a wood way with this

PR #3564, opened 2026-09-04
Fix: deep link handling should popup account select on multi-account …

sonalys:
# Summary
* Add user select behavior before deep-link handling so that the user is capable of selecting which account should handle the link.

image

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [x] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS

### DIFF
--- a/lib/utils/url_launcher.dart
+++ b/lib/utils/url_launcher.dart
@@ -6,6 +6,7 @@
 import 'package:collection/collection.dart' show IterableExtension;
 import 'package:fluffychat/config/app_config.dart';
 import 'package:fluffychat/l10n/l10n.dart';
+import 'package:fluffychat/widgets/adaptive_dialogs/client_picker_dialog.dart';
 import 'package:fluffychat/widgets/adaptive_dialogs/show_ok_cancel_alert_dialog.dart';
 import 'package:fluffychat/widgets/adaptive_dialogs/user_dialog.dart';
 import 'package:fluffychat/widgets/future_loading_dialog.dart';
@@ -129,6 +130,15 @@ class UrlLauncher {
       AppConfig.deepLinkPrefix,
       AppConfig.inviteLinkPrefix,
     );
+
+    // With multiple accounts, let the user choose which account should
+    // handle this link. The chosen client is set as active client so that
+    // all following actions use it.
+    if (matrix.isMultiAccount) {
+      final client = await showClientPickerDialog(context);
+      if (client == null || !context.mounted) return;
+      matrix.setActiveClient(client);
+    }
 
     // The identifier might be a matrix.to url and needs escaping. Or, it might have multiple
     // identifiers (room id & event id), or it might also have a query part.
--- a/lib/widgets/adaptive_dialogs/client_picker_dialog.dart
+++ b/lib/widgets/adaptive_dialogs/client_picker_dialog.dart
@@ -1,0 +1,95 @@
+// SPDX-FileCopyrightText: 2019-Present Christian Kußowski
+// SPDX-FileCopyrightText: 2019-Present Contributors to FluffyChat
+//
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+import 'package:fluffychat/l10n/l10n.dart';
+import 'package:fluffychat/utils/show_scaffold_dialog.dart';
+import 'package:fluffychat/widgets/adaptive_dialogs/adaptive_dialog_action.dart';
+import 'package:fluffychat/widgets/avatar.dart';
+import 'package:fluffychat/widgets/matrix.dart';
+import 'package:material_ui/material_ui.dart';
+import 'package:matrix/matrix.dart';
+
+/// A reusable dialog which lets the user pick one of the logged in
+/// accounts (clients). Returns the selected [Client] or `null` if the
+/// dialog was dismissed.
+///
+/// Shows nothing and returns the current client directly if there is
+/// only a single account.
+Future<Client?> showClientPickerDialog(
+  BuildContext context, {
+  String? title,
+}) async {
+  final matrix = Matrix.of(context);
+  final clients = matrix.widget.clients.where((client) => client.isLogged());
+  if (clients.length < 2) {
+    return clients.isEmpty ? null : matrix.client;
+  }
+  return showScaffoldDialog<Client>(
+    context: context,
+    containerColor: Colors.transparent,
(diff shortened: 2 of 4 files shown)


sonalys:
BTW: `showScaffoldDialog` has some properties that are unused, and it also is not responsive for Desktop screens.
I recommend removing max size and border effects, and only using transparent container color.

lib/utils/url_launcher.dart
```diff
@@ -130,6 +131,15 @@ class UrlLauncher {
       AppConfig.inviteLinkPrefix,
     );
 
+    // With multiple accounts, let the user choose which account should
+    // handle this link. The chosen client is set as active client so that
+    // all following actions use it.
+    if (matrix.isMultiAccount) {
+      final client = await showClientPickerDialog(context);
```
krille-chan, line 138: Can you reuse `showModalActionPopup()` instead? That should make it more minimal

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -17,79 +17,54 @@ import 'package:matrix/matrix.dart';
...
+      }),
     ),
-  );
-}
+  )).result;
 
-class ClientPickerDialog extends StatelessWidget {
-  final List<Client> clients;
-  final String title;
+  if (profiles == null || !context.mounted) return null;
```
sonalys, line 47: Had this unfortunate build gap here that I don't like, but at the same time I don't see a better solution.

krille-chan:
> Can you reuse showModalActionPopup() instead? That should make it more minimal

@sonalys ?

sonalys:
> > Can you reuse showModalActionPopup() instead? That should make it more minimal
> 
> @sonalys ?

@krille-chan  I've implemented using `showModalActionPopup` in https://github.com/krille-chan/fluffychat/pull/3564/changes/516fa6c8bcbc3558fd27ec4da508800ef7c33ed8#diff-3b43a42e400dab16c7c0b48898675797e889a492080a672800a8c723dd651f58

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -0,0 +1,70 @@
...
+/// accounts (clients). Returns the selected [Client] or `null` if the
+/// dialog was dismissed.
+///
+/// Shows nothing and returns the current client directly if there is
+/// only a single account.
+Future<Client?> showClientPickerDialog(BuildContext context) async {
+  final matrix = Matrix.of(context);
+  final clients = matrix.widget.clients.where((client) => client.isLogged());
+
+  if (clients.length < 2) {
```
krille-chan, line 23: you already do this check in url_launcher with `matrix.isMultiAccount`

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -0,0 +1,70 @@
...
+
+/// A reusable dialog which lets the user pick one of the logged in
+/// accounts (clients). Returns the selected [Client] or `null` if the
+/// dialog was dismissed.
+///
+/// Shows nothing and returns the current client directly if there is
+/// only a single account.
+Future<Client?> showClientPickerDialog(BuildContext context) async {
+  final matrix = Matrix.of(context);
+  final clients = matrix.widget.clients.where((client) => client.isLogged());
```
krille-chan: Not needed

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -0,0 +1,70 @@
...
+          final name =
+              profiles[i].displayName ??
+              client.userID?.localpart ??
+              client.userID ??
+              '';
+          return AdaptiveModalAction(
+            label: name,
+            value: client,
+            isDefaultAction: i == 0,
+            icon: profiles[i].avatarUrl != null
```
krille-chan: We should always display the Avatar widget. It has a fallback with just the name

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -0,0 +1,70 @@
...
+    title: L10n.of(context).chooseAccount,
+    cancelLabel: L10n.of(context).cancel,
+    actions: [
+      for (final (i, client) in clients.indexed)
+        () {
+          final name =
+              profiles[i].displayName ??
+              client.userID?.localpart ??
+              client.userID ??
+              '';
```
krille-chan: Empty string is not a good fallback but you can safely assume that we always have a userID

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
sonalys: Are there any downsides towards this redundancy? I particularly like it.

krille-chan: yes, additional lines of code that we not need are always a downside

lib/widgets/adaptive_dialogs/client_picker_dialog.dart
```diff
@@ -0,0 +1,74 @@
...
+import 'package:fluffychat/widgets/matrix.dart';
+import 'package:material_ui/material_ui.dart';
+import 'package:matrix/matrix.dart';
+
+/// A reusable dialog which lets the user pick one of the logged in
+/// accounts (clients). Returns the selected [Client] or `null` if the
+/// dialog was dismissed.
+///
+/// Shows nothing and returns the current client directly if there is
+/// only a single account.
```
krille-chan, line 19: too verbose documentation IMO. Looks like AI also which is prohibited. You have agreed to the contribution guidelines?

lib/utils/url_launcher.dart
```diff
@@ -130,6 +131,15 @@ class UrlLauncher {
       AppConfig.inviteLinkPrefix,
     );
 
+    // With multiple accounts, let the user choose which account should
+    // handle this link. The chosen client is set as active client so that
+    // all following actions use it.
```
krille-chan, line 136: The code is quite self-explainatory. Please remove these comments.

Image

more information why code comments are in 90% of the cases a bad pattern: https://www.youtube.com/watch?v=Bf7vDBBOBUA&t=332s

PR #3575, opened 2026-09-06
fix: guard emoji insertion when the message input was never focused

### LINKED ISSUE #3559: Inserting two emojis from the emojie picker crashes the app when not clicked on the text field before.
### Bug Description

 Inserting two emojis from the picker crashes the app when not clicked on the text field before. It appears a special [?] character before and then app is crashed.

### Steps to Reproduce

1. Open a chat on iOS tap directly the emoji button. 
2. Do NOT tap the text field
4. Tap any emoji. It appears in the field, nothing else happens.
5. Tap a second emoji. -> the app terminates.

### StackTrace or logs

RangeError (start): Invalid value: Not in inclusive range 0..58: -1
#0  RangeError.checkValidRange           (dart:core/errors.dart:387:7)
#1  _StringBase.replaceRange             (dart:core-patch/string_patch.dart:634:33)
#2  ChatController.typeEmoji             (package:fluffychat/pages/chat/chat.dart)
#3  ChatController.onEmojiSelected       (package:fluffychat/pages/chat/chat.dart)
#4  EmojiPickerState._onEmojiSelected    (package:emoji_picker_flutter/src/emoji_picker.dart:388:29)
#5  _DefaultEmojiPickerViewState._onSkinTonedEmojiSelected
#6  EmojiCell.build.onPressed            (package:emoji_picker_flutter/src/widgets/emoji_cell.dart:75:22)
#7  _InkResponseState.handleTap          (package:flutter/src/material/ink_well.dart:1224:21)

### App Version

2.9.3

### Platform

iOS

Device: iPhone 15 Pro Max (iPhone16,2)

### Additional Context

_No response_

ibrahim-iqbal:
Fixes #3559.

### Root cause

When the emoji picker is opened without the message input ever having been focused, `sendController.selection` is `TextSelection.collapsed(offset: -1)` — Flutter's default for an unset selection. In that state:

- On an empty input, the existing `text.isEmpty` shortcut skipped `replaceRange`, so the first emoji appeared to work, but the new selection was set to `-1 + emoji.emoji.length`, leaving the cursor in a half-valid state.
- On the next tap, `text.replaceRange(-1, -1, emoji.emoji)` threw `RangeError (start): Invalid value: Not in inclusive range 0..<text.length>: -1` and crashed the chat page.

### Fix

If either selection endpoint is negative, treat the operation as an append and place the cursor after the inserted glyph. The empty-input shortcut becomes redundant because `text.replaceRange(0, 0, emoji.emoji)` produces the same result, so the branch is removed and the code path is now single.

### Reproduction (from the issue)

1. Open a chat.
2. Tap the emoji button without tapping the text field first.
3. Tap two emojis in a row — previously crashed on the second tap; now both are appended and the cursor sits after the last emoji.


krille-chan:
@ibrahim-iqbal please read and accept the contribution guidelines first. Your pr description looks like ai

krille-chan:
Already found a much smaller fix: https://github.com/krille-chan/fluffychat/pull/3578

PR #3617, opened 2026-09-15
Fix/3425 read marker push rule

inDane:

- [ ] Android
- [ ] iOS
- [ ] Browser (Firefox based)
- [ ] Browser (WebKit based)
- [X] Desktop Linux
- [ ] Desktop Windows
- [ ] Desktop macOS


lib/pages/chat/chat.dart
```diff
@@ -614,21 +614,31 @@ class ChatController extends State<ChatPageWithRoom>
     }
 
     final setOnLatestEvent = eventId == null;
+    // Pick the newest message-type event, independent of push rules.
+    // The pushruleEvaluator only knows account-data rules (the SDK's
+    // PushRuleSet drops the server's default rules and defaults to
+    // notify=false), so plain messages in rooms without custom push rules
+    // never matched and the read marker was silently skipped.
+    // See: https://github.com/krille-chan/fluffychat/issues/3425
```
krille-chan, line 624: Can you please give me clear steps to reproduce this on latest main? I don't have this problem

inDane: TL;DR: ok, mhh, sorry, put this on hold...
I wanted to screenshot what i was doing with a new account etc. aaand i just worked...
With a fresh install/account i was not able to reproduce it. I copied the notification settings too.

---

Damn. This is such an annoying bug. I currently have to get my phone with fluffy chat 2.8.0 to get my channels marked as read, because my linux client 2.9.5 would not mark them... the issue is real, but maybe I looked at the wrong location.
With the client that i've built from source with my patch (on a fresh machine) the channels were marked correctly, so I assumed it was the fix, but now im thinking it might have been the fresh machine that did it.

Anyways, Ill build the regular 2.9.5 from source on a fresh machine and see if that can mark channels read aswell... 

There is something different though. When its working, it is a light colored circle with the number of unread messages next to the channel. When its not working it is a darker colored circle, without a number in it.
```
Light circle + number → notificationCount > 0
Dark circle, no number → hasNewMessages == true but notificationCount == 0 and markedUnread == false 
```

krille-chan: No, please build the main branch from source, no 2.9.5 to reproduce it

inDane: Small status update:
I had the issue today, I opened the main branch built version and the channel was **not** marked read!

Fluffychat_bug_blurred

Using the patched version, the channel was marked read!

I am investigating further...

inDane: Apparently, the bug is connected to these notification settings:
image

i usually have them switched off. Switching them on causes the main branch build to get the channels in "read"-state.

Ill switch them off again and confirm that behavior.

inDane: Yup, these two toggles are the culprit.

krille-chan: I still can not reproduce it or I don't understand it. So I have switched the toggles off:
grafik

and then received a message in a room and entered the room -> was instantly marked as read

I tried it with a muted room, received a message in it and entered the room -> was instantly marked as read

What am I missing?

krille-chan: I was able to reproduce a problem like this:

- Toggle the "Mute all notifications" toggle in the settings
- Receive a message in a **not muted** room
- Enter the room -> NOT marked as read

This also makes sense to me and would need a fix

krille-chan: However I would prefer a different approach than you have in your PR. Your PR now ignores the push rules and just searches for the first message and in a timeline with only state events it would still not work. Will close this in favor of https://github.com/krille-chan/fluffychat/pull/3638

inDane: Thanks for the follow-up!

inDane: i've rebuild from main on commit 3b4f5deaa and my channel was marked read. Looks good, thanks!

